PR #2, opened 2019-05-11
Update README.md

elkimek:
suggesting Tor Relay option

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,7 +46,7 @@ The following services are exposed:
 
 Additional noteworthy components on the BitBox Base:
 
-* [Tor](https://www.torproject.org/): external network connections exclusively use the privacy-focused Tor network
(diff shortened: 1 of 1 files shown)


Stadicus:
Thanks for the suggestion to support the Tor network. We discussed the impact of our appliance on the Tor network as well and think it's an important issue. I think that it's not as easy as to enable the relay option,however, as this is a delicate feature if users don't understand the consequences.

We continue to look into option on how to minimize negative externalities.

Stadicus:
At the current stage of the project, we don't want to promise running the BitBox Base as a Tor relay. I could imagine having it as an option in the "Advanced features", though. Also, me could add a VPN option for IBD later, but not for the MVP.

We won't have a major impact on the Tor network at the moment, as we are shipping with presynced blockchain data and a full validation is possible without redownloading everything.

PR #5, opened 2019-05-13
build.sh: easier adjustment of VirtualBox settings

bavarianledger:
added a #Settings section into build.sh to allow easier adjustments of the VirtualBox settings (number of CPUs, memory)

### DIFF
--- a/armbian/build.sh
+++ b/armbian/build.sh
@@ -7,6 +7,13 @@
 #
 set -eu
 
+# Settings
+#
+# VirtualBox Number of CPU cores
+VIRTUALBOX_CPU="4"
+# VirtualBox Memory in MB
+VIRTUALBOX_MEMORY="8192"
+
 function usage() {
 	echo "Build customized Armbian base image for BitBox Base"
 	echo "Usage: $0 [update]"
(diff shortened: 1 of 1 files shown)


armbian/build.sh
```diff
@@ -7,6 +7,13 @@
 #
 set -eu
 
+# Settings
+#
+# VirtualBox Number of CPU cores
+VIRTUALBOX_CPU="4"
```
hkjn: (Optional) We could also take the environment variable from the caller, and setting it to the current value in case it's unspecified:

```
VIRTUALBOX_CPU=${VIRTUALBOX_CPU:-"4"}
```

That way, an invocation like `VIRTUALBOX_CPU=8 bash build.sh` would e.g use 8 cores.

(Same suggestion for `VIRTUALBOX_MEMORY`.)

hkjn:
I'll copy over the suggestion I made in a now-closed thread to not lose track of it:
> (Optional) We could also take the environment variable from the caller, and setting it to the current value in case it's unspecified:
> 
> ```
> VIRTUALBOX_CPU=${VIRTUALBOX_CPU:-"4"}
> ```
> 
> That way, an invocation like `VIRTUALBOX_CPU=8 bash build.sh` would e.g use 8 cores.
> 
> (Same suggestion for `VIRTUALBOX_MEMORY`.)

I tried writing this up myself as a follow-up PR, but noticed that there's some issues that makes it less trivial than it appears; passing through the variable from the environment is easy, but we currently only apply the `sed` to edit `armbian-build/Vagrantfile` the first time we do `git clone https://github.com/armbian/build`, so subsequent changes to the environment variable won't affect the size of the Vagrant VM used.

Trying to change this and also handle both the case where `#vb.memory` is commented-out or not leads to a need for a `sed` expression that is beyond my powers to write quickly. Better yet would be to not have to use the `sed` at all, in case there's some more flexible mechanism to override Vagrant settings. I've left some trails on a branch below for posterity, but don't have the time to look into it further myself at this time:

https://github.com/digitalbitbox/bitbox-base/compare/master...hkjn:20190513-passthrough-params?expand=1

bavarianledger:
Yes, I noticed that there are a couple of general design issues that have to be considered, like the `sed`-issue, environment variables, behaviour of `build.sh update` or putting the settings into an `include file`.
I'll leave this to you until you have decided how you want to proceed ;)

PR #6, opened 2019-05-13
avahi-daemon: publish-workstation=yes

Stadicus:
Configure the `avahi-daemon` to publish the hostname within the local network and make the device reachable, e.g. under `bitbox-base.local`

### DIFF
--- a/armbian/base/build/customize-armbian-rockpro64.sh
+++ b/armbian/base/build/customize-armbian-rockpro64.sh
@@ -700,6 +700,8 @@ Cache=yes
 EOF
 
 # mDNS services
+sed -i '/PUBLISH-WORKSTATION/Ic\publish-workstation=yes' /etc/avahi/avahi-daemon.conf
+
 cat << EOF > /etc/avahi/services/bitcoind.service
 <?xml version="1.0" standalone='no'?>
 <!DOCTYPE service-group SYSTEM "avahi-service.dtd">


hkjn:
[The docs](https://linux.die.net/man/5/avahi-daemon.conf) seem to say the following on the `publish-workstation` config value:

> publish-workstation= Takes a boolean value ("yes" or "no"). If set to "yes" avahi-daemon will register a service of type "_workstation._tcp" on the local LAN. This might be useful for administrative purposes (i.e. browse for all PCs on the LAN), but is not required or recommended by any specification. Newer MacOS X releases register a service of this type. Defaults to "yes".

On the local LAN setup at Shift HQ, it seems that the node with hostname `alice` already resolves to the internal IPv4 address, both when specified as `alice` and `alice.local` and `alice.localdomain`.

@Stadicus: Could you clarify a) what this setting does, beyond what's already working and b) whether the `Defaults to "yes"` part of the docs is inaccurate?

Stadicus:
This option results in a dedicated mDNS broadcast, visible when scanning the network (I named my node "frank"):
```
$ avahi-browse -a
+ wlp59s0 IPv6 frank [ce:d0:c8:e5:e8:c7]                     Workstation          local
+ wlp59s0 IPv4 frank [ce:d0:c8:e5:e8:c7]                     Workstation          local
```

Without this setting, these entries are no longer broadcasted. I know that our router in the office already takes care of that, but that's not necessarily the case with every router configuration.

Let's wait before merging, I'll ask some people with network setups that do not broadcast that automatically.

hkjn:
Thanks @Stadicus, that all makes sense to me. Happy to merge this as-is if it solves problems on other networking setups.

We could consider adding some of this context to the commit messages, so it's available for future reference.

utACK 0d4ddbea96f5c4b43ebe54fcbd051ad0d0d642e8

PR #8, opened 2019-05-13
armbian/base: Switch to cookie based auth

hkjn:


armbian/base/build/customize-armbian-rockpro64.sh
```diff
@@ -257,6 +255,11 @@ After=network.target startup-checks.service
 Requires=startup-checks.service
 [Service]
 ExecStart=/usr/bin/bitcoind -daemon -conf=/etc/bitcoin/bitcoin.conf 
+# Note: we need this ExecStartPost command to prepare an .env file
+# in expected format for electrs to have the RPCPASSWORD in its
+# environment. If we contributed a --cookie-file flag to upstream
+# electrs that could parse the .cookie file, we would not need this.
+ExecStartPost=/bin/bash -c "sleep 5 && echo \"RPCPASSWORD=$(tail -c +12 /mnt/ssd/bitcoin/.bitcoin/testnet3/.cookie)\" > /mnt/ssd/bitcoin/.bitcoin/testnet3/.cookie.env"
```
Stadicus: Loading the content of the bitcoind cookie file from standard location would mean that all locations have to be adjusted when switching from testnet to mainnet. This means that the script `base/scripts/set-bitcoin-network.sh` needs to be extended as well.

Instead, I propose to store the `.cookie` always in the same location, regardless of the network that currently is active. For example with this configuration in `/etc/bitcoin/bitcoin.conf`:
```
rpccookiefile=/mnt/ssd/bitcoin/.bitcoin/.cookie
```

hkjn: Great idea, done.

hkjn:
Thanks @Stadicus, PTAL.

I made the change you requested to always use the same path to the `.cookie` file, and also made the `ExecStartPost` command more readable.

I also made the `RestartSec` 60 sec for `bitcoind.service` since it seems that the shutdown + start took more than 10 sec at times in my testing, and I made the `electrs.service` have a hard `Requires=bitcoind.service` (earlier it was `Wants=bitcoind.service`) to ensure that a restart of `bitcoind.service` also triggers a restart of `electrs.service`, so the latter has the most recent credentials from the `.cookie`.

Stadicus:
**TL;DR: it's a typo, see the very end...**

I first tried to configure my existing Base with the script (should work) but `bitcoind` would not start afterwards. Then I built the whole system from scratch and swapped the eMMC, but with the same result.

It seems that the startup process is killed after 15 seconds.

```
2019-05-13T20:34:46Z Bitcoin Core version v0.18.0 (release build)
...
2019-05-13T20:34:46Z Using data directory /mnt/ssd/bitcoin/.bitcoin
2019-05-13T20:34:46Z Config file: /etc/bitcoin/bitcoin.conf
2019-05-13T20:34:46Z No rpcpassword set - using random cookie authentication.
2019-05-13T20:34:46Z Generated RPC authentication cookie /mnt/ssd/bitcoin/.bitcoin/.cookie
...
2019-05-13T20:34:56Z Loaded best chain: hashBestChain=0000000000000000002920fbe483b2c7f727af15bfe0303a74818fbfe163a4e3 height=575903 date=2019-05-13T19:58:53Z progress=0.999981
2019-05-13T20:34:56Z init message: Rewinding blocks...
2019-05-13T20:34:59Z init message: Verifying blocks...
2019-05-13T20:34:59Z Verifying last 6 blocks at level 3
2019-05-13T20:34:59Z [0%]... block index           13396ms
2019-05-13T20:34:59Z Shutdown requested. Exiting.
2019-05-13T20:34:59Z Shutdown: In progress...
2019-05-13T20:34:59Z scheduler thread interrupt
2019-05-13T20:35:00Z Shutdown: done
```

journald says the following:
```
May 13 22:34:46 frank-hkjn systemd[1]: Starting Bitcoin daemon...
May 13 22:34:51 frank-hkjn bash[2659]: tail: cannot open '/mnt/bitcoin/.bitcoin/.cookie' for reading: No such file or directory
May 13 22:34:51 frank-hkjn systemd[1]: bitcoind.service: Control process exited, code=exited status=1
May 13 22:35:01 frank-hkjn systemd[1]: Failed to start Bitcoin daemon.
```

So even though `bitcoind` says at 20:34:46 (UTC) that it generated the cookie, the ExecStartPost runnint `tail` cannot find it. It cannot be a permission problem as it's the same user.

After questioning the whole thing, I found a trivial typo:
```
this      --> tail -c +12 /mnt/bitcoin/.bitcoin/.cookie
should be --> tail -c +12 /mnt/ssd/bitcoin/.bitcoin/.cookie
```

armbian/base/build/customize-armbian-rockpro64.sh
```diff
@@ -257,13 +256,21 @@ After=network.target startup-checks.service
...
 [Service]
 ExecStart=/usr/bin/bitcoind -daemon -conf=/etc/bitcoin/bitcoin.conf 
+# Note: we need this ExecStartPost command to prepare an .env file
+# in expected format for electrs to have the RPCPASSWORD in its
+# environment. If we contributed a --cookie-file flag to upstream
+# electrs that could parse the .cookie file, we would not need this.
+ExecStartPost=/bin/bash -c " \
+  sleep 5 && \
+  echo -n "RPCPASSWORD=" > /mnt/ssd/bitcoin/.bitcoin/.cookie.env && \
+  tail -c +12 /mnt/bitcoin/.bitcoin/.cookie >> /mnt/ssd/bitcoin/.bitcoin/.cookie.env"
```
Stadicus: > `tail -c +12 /mnt/bitcoin/.bitcoin/.cookie`

typo: should be `tail -c +12 /mnt/ssd/bitcoin/.bitcoin/.cookie`

armbian/base/build/customize-armbian-rockpro64.sh
```diff
@@ -257,13 +256,21 @@ After=network.target startup-checks.service
 Requires=startup-checks.service
 [Service]
 ExecStart=/usr/bin/bitcoind -daemon -conf=/etc/bitcoin/bitcoin.conf 
+# Note: we need this ExecStartPost command to prepare an .env file
+# in expected format for electrs to have the RPCPASSWORD in its
+# environment. If we contributed a --cookie-file flag to upstream
+# electrs that could parse the .cookie file, we would not need this.
+ExecStartPost=/bin/bash -c " \
+  sleep 5 && \
```
Stadicus: maybe increase to 10s or more, just to be safe, as it takes some time until the cookie file is written.

Stadicus:
Additional issue: both `lightningd.service` and `electrs.service` fail on startup. I noticed that with lightningd before, but not with electrs. That worked some days ago, I need to check that tomorrow.

```
May 13 22:59:38 frank-hkjn systemd[1]: Dependency failed for c-lightning daemon.
May 13 22:59:38 frank-hkjn systemd[1]: lightningd.service: Job lightningd.service/start failed with result 'dependency'.

May 13 22:59:38 frank-hkjn systemd[1]: Dependency failed for Electrs server daemon.
May 13 22:59:38 frank-hkjn systemd[1]: electrs.service: Job electrs.service/start failed with result 'dependency'.
```

Stadicus:
Additional issue (sorry!): on testnet, c-lightning expects the .cookie file to be present in the `.bitcoin/testnet3` directory. Currently it does not start.

I see two options:
1) test if the lightningd argument `--bitcoin-datadir <arg>` works for the .cookie file also, as there is no dedicated argument 
1) create a symlink that stays there always (works if .cookie is present, and is broken if it isn't). That needs to be included in `startup-checks.sh`:  
    ```
    mkdir -p /mnt/ssd/bitcoin/.bitcoin/testnet3/
    ln -s /mnt/ssd/bitcoin/.bitcoin/.cookie /mnt/ssd/bitcoin/.bitcoin/testnet3/.cookie
    ```

armbian/base/build/customize-armbian-rockpro64.sh
hkjn: Oops, good catch.

Thanks, fixed.

armbian/base/build/customize-armbian-rockpro64.sh
hkjn: Done.

hkjn:
> Additional issue (sorry!): on testnet, c-lightning expects the .cookie file to be present in the .bitcoin/testnet3 directory. Currently it does not start.

Ah, due to the `rpccookiefile` change, got it. I went with option number two and updated `startup-checks.sh` to create a symlink.

Thanks Roland, PTAL.

hkjn:
>>    Additional issue: both lightningd.service and electrs.service fail on startup.
>
> I'm not seeing this on my Base node running on testnet. I'll rebuild from the latest commit and try again. Did you try on mainnet or testnet?

After rebuilding the image, I'm also seeing this behavior:
```
May 14 11:55:09 bitbox-base systemd[1]: Dependency failed for c-lightning daemon.
May 14 11:55:09 bitbox-base systemd[1]: lightningd.service: Job lightningd.service/start failed with result 'dependency'.
```

Still investigating what changed in regards to `bitcoind.service`/`lightningd.service` interaction..

hkjn:
> > > Additional issue: both lightningd.service and electrs.service fail on startup.
> > 
> > 
> > I'm not seeing this on my Base node running on testnet. I'll rebuild from the latest commit and try again. Did you try on mainnet or testnet?
> 
> After rebuilding the image, I'm also seeing this behavior:
> 
> ```
> May 14 11:55:09 bitbox-base systemd[1]: Dependency failed for c-lightning daemon.
> May 14 11:55:09 bitbox-base systemd[1]: lightningd.service: Job lightningd.service/start failed with result 'dependency'.
> ```
> 
> Still investigating what changed in regards to `bitcoind.service`/`lightningd.service` interaction..

We (@Stadicus and I) looked into this together and found that these dependency ordering issues also seem to be present in the `master` branch. I've added `ExecStartPre=/bin/sleep 5` now as a workaround to this PR, and a comment explaining what's going on and that we should look into a better solution.

> Ah, due to the rpccookiefile change, got it. I went with option number two and updated startup-checks.sh to create a symlink.

I noticed that the `ln -s` command also needs `-f`, or re-running `startup-checks.service` will cause it to fail due to the existing symlink target. I've fixed the command to include `-f` at this point as well.

hkjn:
Okay, this PR should at last be ready for review again. We hit some interesting systemd behavior in the form of https://github.com/systemd/systemd/issues/1312, but have added workarounds that seems to consistently produce the behavior we want (i.e all units start successfully), without the fragility of the earlier `sleep 5` hack.

PTAL.

Stadicus:
utACK https://github.com/digitalbitbox/bitbox-base/commit/14cc525c81c2acd00c430022daa5ced547c96892

Stadicus:
Built image and all services come up reliably. I still don't know why `systemd` does not respect the `After=bitcoin.service` in the lightningd & electrs units, as they also get started immediately. My expectation would be that they hold off until the `ExecStartPost=` is finished as well.

hkjn:
> Built image and all services come up reliably. I still don't know why `systemd` does not respect the `After=bitcoin.service` in the lightningd & electrs units, as they also get started immediately.

I could be wrong here, but I'd expect that if we have `b.service` that specifies `After=a.service`, we'd see `b.service` start up after the `ExecStart` line of `a.service` has come up, i.e at the time where `systemctl is-active a.service` returns `active`.

PR #13, opened 2019-05-16
armbian/base: Add basic iptables.rules config

hkjn:
This is a fairly restrictive networking config that disallows
all ingress and egress traffic apart from a known good set of
services. The functionality of the Base appears unharmed, with
the grafana dashboard loading fine, all services appearing to
continue running, no obvious errors and no blocked traffic
(which appears in `journalctl --all -f` logs) that seems unintentional.

This should be integrated into the systemd services to automatically
do `iptables-restore < <path to iptables.rules>` after the network
comes up and before any services are started.

### DIFF
--- a/armbian/base/config/iptables/iptables.rules
+++ b/armbian/base/config/iptables/iptables.rules
@@ -1,0 +1,130 @@
+*filter
+# Default policy is DROP; we need to explicitly allow any packets we
+# do want.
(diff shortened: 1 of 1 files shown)


Stadicus:
tACK https://github.com/digitalbitbox/bitbox-base/commit/b9187998d2b181075e7165c2cae0a6985ef410e0

PR #15, opened 2019-05-20
Freeze Armbian & additional build options

Stadicus:
Various minor fixes and features adds
* freeze Armbian at latest stable release (sunxi-4.20)
* build option to manually set root password (development only)
* build option to configure default wifi network (needs additional hardware module)
* display build options (dedicated script function)


Stadicus:
To be replaced by rebased branch / pr

PR #16, opened 2019-05-21
Wifi configuration on build & Armbian freeze

Stadicus:
Various minor fixes and features adds
* freeze Armbian at latest stable release (sunxi-4.20)
* build option to manually set root password (development only)
* build option to configure default wifi network (needs additional hardware module)
* display build options (dedicated script function)

### DIFF
--- a/armbian/base/build/customize-armbian-rockpro64.sh
+++ b/armbian/base/build/customize-armbian-rockpro64.sh
@@ -28,6 +28,18 @@
 
 set -e
 
+function echoArguments {
+  echo "
+================================================================================
+==> $1
+================================================================================
+    USER / PASSWORD: root / ${BASE_ROOTPW}
+    HOSTNAME:        ${BASE_HOSTNAME}
+    BITCOIN NETWORK: ${BASE_BITCOIN_NETWORK}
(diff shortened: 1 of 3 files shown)


armbian/base/build/customize-armbian-rockpro64.sh
```diff
@@ -740,6 +758,22 @@ DNSSEC=yes
 Cache=yes
 EOF
 
+cat << EOF > /etc/network/interfaces
```
hkjn: Do we need to override `/etc/network/interfaces`? On the `eve` node, built from a recent commit on `master`, it seems to already exist and have the equivalent content as the file you're creating here:

```
root@eve:~# cat /etc/network/interfaces
source /etc/network/interfaces.d/*
# Network is managed by Network manager
auto lo
iface lo inet loopback
```

Stadicus: I'll try if it just works like that.

Stadicus: Removed the explicit `/etc/network/interfaces`, works just fine.

hkjn:
Post-merge utACK 84a201063221bb340b73ad36e6d2bcc9108fad52

PR #20, opened 2019-05-22
bbb-config: general system configuration utility

Stadicus:
**bbb-config**
* bbb-config.sh: new general system configuration utility
* new key/value pair configuration store at /opt/shift/sysconfig
* adjust startup-checks
* rename systemd-base.sh to bbb-systemctl.sh
* add alias in .bashrc: bbbconfig & bbbsystemctl
* deprecate set-bitcoin-network.sh
* rename build option to BASE_WIFI_PW
* documentation

**HDMI dashboard**
* install local x-server & chromium
* add user "hdmi" for HDMI dashboard
* configure openbox and startx
* _autologin of user "hdmi" not yet implemented_


Stadicus:
Draft, needs final build and testing, review is appreciated

armbian/base/build/customize-armbian-rockpro64.sh
```diff
@@ -107,6 +107,7 @@ adduser --system --group --disabled-login --no-create-home              electrs
 adduser --system --group --disabled-login --home /var/run/avahi-daemon  avahi
 adduser --system --group --disabled-login --no-create-home              prometheus
 adduser --system --group --disabled-login --no-create-home              node_exporter
+adduser --disabled-password --gecos "" hdmi
```
hkjn: I wasn't familiar with the `--gecos` flag. I take it this is to avoid needing to specify user details (name, email, "room number") for a non-system user? Speaking of which, is there a reason `hdmi` shouldn't / couldn't be a `--system` user?

armbian/base/build/customize-armbian-rockpro64.sh
```diff
@@ -168,6 +169,8 @@ cat << EOF > /root/.bashrc-custom
 export LS_OPTIONS='--color=auto'
 alias l='ls $LS_OPTIONS -l'
 alias ll='ls $LS_OPTIONS -la'
+alias bbbconfig='bbb-config.sh'
```
hkjn: (Optional) Personally I probably wouldn't add these aliases, since they're not that much shorter than the name of the scripts they are aliasing anyway, and it seems like having more than one way to do the same thing could lead to more confusion.

armbian/base/build/customize-armbian-rockpro64.sh
```diff
@@ -200,6 +203,10 @@ EOF
 # retain journal logs between reboots 
 ln -sf /mnt/ssd/system/journal/ /var/log/journal
 
+# SYSTEM CONFIGURATION ---------------------------------------------------------
+SYSCONFIG_PATH="/opt/shift/sysconfig/"
+mkdir -p $SYSCONFIG_PATH
+echo "BITCOIN_NETWORK=testnet" > ${SYSCONFIG_PATH}BITCOIN_NETWORK
```
hkjn: (Optional) If we omitted the trailing `/` in `SYSCONFIG_PATH` on L207, we'd get a more natural expression for the full path here:
```
echo "BITCOIN_NETWORK=testnet" > ${SYSCONFIG_PATH}/BITCOIN_NETWORK
```

hkjn: I would call the file we write to `bitcoin_network.env` or `bitcoin_network.conf` or similar, to a) give a hint through the file suffix that the file is a text file for environment variables / configuration and b) use lower case.

(And elsewhere.)

armbian/base/build/customize-armbian-rockpro64.sh
```diff
@@ -661,6 +668,7 @@ EOF
 
 # NGINX ------------------------------------------------------------------------
 apt install -y nginx
+rm /etc/nginx/sites-enabled/default || true
```
hkjn: By using `rm -f`, the command would exit success (`0`) even if the target file did not exist, and we could avoid the `|| true`.

armbian/base/build/customize-armbian-rockpro64.sh
hkjn: Should adding the `hdmi` user, installing `chromium` etc be conditional, and disabled by default, maybe? That way, we'd avoid making the image / attack surface larger for users that don't need this feature.

armbian/base/build/customize-armbian-rockpro64.sh
```diff
@@ -759,12 +796,10 @@ EOF
 
 # include Wifi credentials, if specified
 if [[ -n "${BASE_WIFI_SSID}" ]]; then
-  cat << EOF > /etc/network/interfaces.d/wlan0.conf
-auto wlan0
-iface wlan0 inet dhcp
-  wpa-ssid ${BASE_WIFI_SSID}
-  wpa-psk ${BASE_WIFI_PASS}
-EOF
+  sed -i '/WPA-SSID/Ic\  wpa-ssid ${BASE_WIFI_SSID}' /opt/shift/config/wifi/wlan0.conf
```
hkjn: Shouldn't we be replacing the lower-case `wpa-ssid` here? (Not `WPA-SSID`.)

(And elsewhere.)

armbian/base/config/wifi/wlan0.conf
```diff
@@ -0,0 +1,4 @@
+auto wlan0
+iface wlan0 inet dhcp
+  wpa-ssid XXXXXX
```
hkjn: (Optional) We could also use something like `{wpa-password}` as a placeholder value, to more clearly mark that this is not a credential but a place where a script should insert a credential.

C.f how Shesek updates placeholders in `spark-wallet.service`:
* https://github.com/shesek/spark-wallet/blob/master/doc/startup-systemd.md#adding-spark-to-startup-with-systemd

(And elsewhere.)

armbian/base/scripts/bbb-config.sh
```diff
@@ -0,0 +1,210 @@
+#!/bin/bash
+set -e
```
hkjn: Maybe we should also set `-u`, to catch undefined variables being used?

(And elsewhere.)

armbian/base/scripts/bbb-config.sh
```diff
@@ -0,0 +1,210 @@
...
+            ENABLE=1
+        else
+            ENABLE=0
+        fi
+
+        case "${SETTING}" in
+            DASHBOARD_HDMI)
+                # enable / disable auto-login for user "hdmi", start / kill xserver
+                if [[ $ENABLE -eq 1 ]]; then
+                    # todo
```
hkjn: If this is intended to be merged like this, could we add the person who can provide more context in the TODO comment, and write what the comment entails? I.e:

```
TODO(Stadicus): Implement this.
```

armbian/base/scripts/bbb-config.sh
```diff
@@ -0,0 +1,210 @@
...
+        if [[ "${COMMAND}" == "enable" ]]; then
+            ENABLE=1
+        else
+            ENABLE=0
+        fi
+
+        case "${SETTING}" in
+            DASHBOARD_HDMI)
+                # enable / disable auto-login for user "hdmi", start / kill xserver
+                if [[ $ENABLE -eq 1 ]]; then
```
hkjn: If we're using curly braces (`${VAR}`), I think we should do it consistently, i.e `${ENABLE}`.

(And elsewhere.)

armbian/base/scripts/set-bitcoin-network.sh
```diff
@@ -5,7 +5,7 @@ set -e
 # 
 
 function usage() {
-  echo "BitBox Base: set Bitcoin network"
+  echo "BitBox Base: set Bitcoin network (DEPRECATED)"
```
hkjn: Do we really need to deprecate this script? Couldn't we just remove it, to avoid the overhead of maintaining two scripts for some time?

armbian/base/build/customize-armbian-rockpro64.sh
Stadicus: When using the `--system` flag, a system user is created non-interactively, with no login shell and a UID < 1000. As the user "hdmi" needs to log in to start the x-server, I decided to create a regular user with this `--gecos` flag (which prevents interactive prompts).

armbian/base/build/customize-armbian-rockpro64.sh
Stadicus: I could live with the `.env` suffix, although that makes it not as clean to write and source the values in my opinion.

Why would you prefer the lowercase filename if the containing variable is uppercase?

armbian/base/build/customize-armbian-rockpro64.sh
Stadicus: Adding this functionality should be done when building the image, and I think a flag to enable / disable makes sense (just not a high priority).

I don't think the attack surface is higher if the auto-login for the "hdmi" user (that has no password) is disabled and the x-server is not running, though. So it's more a question of ressource use in my opinion.

armbian/base/build/customize-armbian-rockpro64.sh
Stadicus: I consistently `sed` for case-insensitive "UPPER CASE" search strings, also in other scripts. This way, if it works, I'm sure it's really not case-sensitive.

armbian/base/config/wifi/wlan0.conf
Stadicus: Interesting, very focused `sed`, good to know. I still prefer to search for a keyword and replace the whole line. I can use the placeholders anyway, to make their nature more clear.

armbian/base/scripts/bbb-config.sh
Stadicus: Good point. I don't intend to merge it in this state, though, that's why the pull request is a draft.

armbian/base/scripts/set-bitcoin-network.sh
Stadicus: Definitely, just wanted to test it first. Removed.

armbian/base/build/customize-armbian-rockpro64.sh
hkjn: For my education, could you explain why having a system user with UID < 1000 prevents starting `X`?

Regardless, perhaps we could add a comment explaining why `--gecos ""` is specified, so that future maintainers will have the context for why we set up the `hdmi` user in this way?

armbian/base/build/customize-armbian-rockpro64.sh
hkjn: > I could live with the .env suffix, although that makes it not as clean to write and source the values in my opinion.

I'm not sure I understand what makes it less clean to do e.g `source bitcoin_network.env`, instead of `source bitcoin_network`.

> Why would you prefer the lowercase filename if the containing variable is uppercase?

UNIX style file naming usually uses lower-case filenames. I don't believe there's any reason why the file name's case has to match the contents of that file; e.g `.bashrc` is lower-case, even if it might set/modify environment variables with upper-case names, like `PATH` for example.

hkjn: > Why would you prefer the lowercase filename if the containing variable is uppercase?

Ah, now I see what you were going for — you define a single variable in each file, instead of a set of variables related to a certain service or piece of infrastructure (e.g Bitcoin, or networking).

Sure, if you prefer to have `$FOO` defined in the file `FOO` and to keep a 1:1 mapping between file containing environment variables and the environment variables defined within, I won't object. I'm not sure it's a very common way to manage configuration, but as long as we stay consistent and it works well for you, go ahead.

armbian/base/scripts/bbb-config.sh
hkjn: Cool, I wasn't clear which parts you wanted me to review or not, which is why I asked. As long as we either implement this or add more context to the TODO comment before merging, that's fine.

armbian/base/build/customize-armbian-rockpro64.sh
Stadicus: I don't really care either way, I just thought a user that logs in should not be a system user. One could argue that "hdmi" is a system user, albeit its login capability. 

On second thought I'll switch it to system user to have a more uniform way of creating the users.  Just need to configure the shell using `chsh -s /bin/bash hdmi`.

PR #24, opened 2019-05-25
Switch from "root" user to sudo user "base"

Stadicus:
This pull requests increases operational security by
* creating different user groups: `bitcoin` for sensitive data, `system` for regular background processes
* creates the sudo user `base` for ssh or local login
* locks the user `root` for regular login
* provides build configuration options to completely disallow password authentication and / or ssh access for user "root"
* remove "world" access rights (r/w/x) from `/mnt/ssd` completely (relevant for user "hdmi")
* installs `fail2ban`

Further changes:
* add Tor hidden service for Middleware
* fix links in documentation
* minor fixes


armbian/base/build/customize-armbian-rockpro64.sh
```diff
@@ -73,57 +78,70 @@ fi
...
-echo "Configured for Bitcoin TESTNET"; echo
-EOF
-chmod 755 /etc/update-motd.d/20-shift
 
-echo "${BASE_HOSTNAME}" > /etc/hostname
-hostname -F /etc/hostname
+# USERS & LOGIN-----------------------------------------------------------------
+# - group 'bitcoin' covers sensitive information
+# - group 'system' is used for service users without sensitive privileges
+# - user 'root' is locked for login
```
hkjn: (Optional) Perhaps "disabled from logging in with password" would be more clear than "locked for login"?

Alternatively, if the aim is to disable the `root` account, perhaps we should follow the recommendations of `man passwd`:
```
           Note that this does not disable the account. The
           user may still be able to login using another
           authentication token (e.g. an SSH key). To
           disable the account, administrators should use
           usermod --expiredate 1 (this set the account's
           expire date to Jan 2, 1970).
```

armbian/base/build/customize-armbian-rockpro64.sh
```diff
@@ -73,57 +78,70 @@ fi
...
+# add groups
+addgroup --system bitcoin
+addgroup --system system
+
+# Set root password (either from configuration or random) and lock account
+BASE_ROOTPW=${BASE_ROOTPW:-$(< /dev/urandom tr -dc A-Z-a-z-0-9 | head -c32)}
+echo "root:${BASE_ROOTPW}" | chpasswd
+passwd -l root
+
+# add user 'base' to sudo and other groups (with options for non-interactive cmd)
```
hkjn: (Optional) Perhaps `--gecos "" is used to prevent interactive prompting for user information` or similar would be more clear than `with options for non-interactive cmd`)?

armbian/base/build/customize-armbian-rockpro64.sh
```diff
@@ -182,7 +200,28 @@ EOF
 
 
 # OS CONFIG --------------------------------------------------------------------
-cat << EOF > /root/.bashrc-custom
+# customize MOTD
+echo "MOTD_DISABLE='header tips updates armbian-config'" >> /etc/default/armbian-motd
+cat << EOF > /etc/update-motd.d/20-shift
+#!/bin/bash
+. /etc/os-release
```
hkjn: (Optional) I find it more readable to use `source` instead of `.`:

```
source /etc/os-release
source /etc/armbian-release
```

armbian/base/build/customize-armbian-rockpro64.sh
```diff
@@ -218,31 +257,41 @@ EOF
 # retain journal logs between reboots 
 ln -sf /mnt/ssd/system/journal/ /var/log/journal
 
+# make bbb scripts executable by sudo
```
hkjn: Nit: I think `with sudo` would be more clear than `by sudo`. (`sudo` is not a user, but a tool to execute commands as the super-user.)

armbian/base/build/customize-armbian-rockpro64.sh
```diff
@@ -392,18 +441,18 @@ ELECTRS_VERSION="0.6.1"
 # ./install.sh
 #
 # apt install clang cmake
-# git clone https://github.com/romanz/electrs
+# git clone https://github.com/romanz/electrs || true
```
hkjn: Nit: Do we need to add the `|| true` to a commented-out command?

Actually, do we need to keep the commented-out commands in the script in the first place?

armbian/base/scripts/bbb-systemctl.sh
```diff
@@ -6,24 +6,26 @@ set -eu
...
 }
 
 ACTION=${1:-"status"}
 
-if ! [[ ${ACTION} =~ ^(status|start|restart|stop|enable|disable)$ ]]; then
-	usage
-	exit 1
+if [[ ${ACTION} == "-h" ]] || [[ ${ACTION} == "--help" ]]; then
+  usage
+	exit 0
```
hkjn: Nit: Mixed indentation (hard tab). We should use the same indentation consistently, which seems to be two spaces for `.sh` scripts.

(And elsewhere.)

armbian/base/build/customize-armbian-rockpro64.sh
Stadicus: I am planning to clean up the overall build (also remove unused packages at the end) in a seperate pull request.

armbian/base/build/customize-armbian-rockpro64.sh
Stadicus: For the moment I'd leave root login with ssh-keys enabled by default (it still can be disabled with the build option `BASE_SSH_ROOT_LOGIN=false`). The danger of locking oneself out, e.g. because the sudo password is unknown, is quite big.

armbian/base/build/customize-armbian-rockpro64.sh
Stadicus: That's the Armbian standard, but I completely agree. Changed to `source`.

PR #25, opened 2019-05-27
Run iptables-restore as systemd unit on startup

Stadicus:
Apply the iptables firewall settings on startup with a systemd service unit.


armbian/base/build/customize-armbian-rockpro64.sh
```diff
@@ -917,6 +917,19 @@ cat << 'EOF' > /etc/avahi/services/bitboxbase.service
 </service-group>
 EOF
 
+# firewall: restore iptables rules on startup
+cat << 'EOF' > /etc/systemd/system/iptables-restore.service
+[Unit]
+Description = BitBox Base: restore iptables rules
```
hkjn: Nit: Maybe we could be consistent with the spaces for systemd values in unit definitions? If we stay with the same style as elsewhere, this would be:
```
Description=BitBox Base: restore iptables rules
```

PR #26, opened 2019-05-28
Add .gitattributes

NicolasDorier:
This makes sure Windows does not mess up add during checkout, this would make the scripts fail.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,5 @@
+# Declare files that will always have CRLF line endings on checkout.
+*.sh text eol=lf
+*.py text eol=lf
+*.service text eol=lf
+Makefile text eol=lf


.gitattributes
```diff
@@ -0,0 +1,7 @@
+# Declare files that will always have CRLF line endings on checkout.
+*.sh text eol=lf
+*.py text eol=lf
+*.service text eol=lf
+Makefile text eol=lf
+configure text eol=lf
```
hkjn: (Optional) I don't believe we have `configure` or `.gitmodules` files in this repo.. is it necessary to list them anyway? Or do you use the same `.gitattributes` file across different projects and would prefer not to modify it?

NicolasDorier: I just removed it, I copy pasted from another project.

NicolasDorier:
More will come. Building dockerfile to easily build all of this stuff.

hkjn:
> More will come.

Great to hear!

> Building dockerfile to easily build all of this stuff.

Interesting. Are you referring to the Armbian image build process, or something else?

Since the Armbian image is built inside a VirtualBox VM, which is managed with Vagrant, I thought that containerizing it might be difficult... then I did a quick search, and found that there are crazy people (in the most positive sense) that have gotten VirtualBox-inside-Docker working:

* https://github.com/jessfraz/dockerfiles/blob/master/virtualbox/Dockerfile

However, my understanding is that you'd a) you'd need to run copy out kernel modules to the host and b) give `--privileged` access to the host when running the container, so I'm not sure what exactly the benefits would be?

I spoke to @Stadicus briefly offline, and he mentions that we think we'd likely want to stick to the Armbian build process (which is where the Vagrant + VirtualBox setup comes from) for the time being.

If we're instead talking about dockerizing the build process of tools like the [bbbfancontrol.go](https://github.com/digitalbitbox/bitbox-base/blob/master/tools/bbbfancontrol/bbbfancontrol.go), that sounds like it likely would be an improvement.

You might also want to be aware of / follow PR #28, which reworks the way the build inputs for the Armbian image are compiled, and which should be merged soon.

NicolasDorier:
@hkjn docker build is supported by armbian. If I manage to make it work (big `if`), I am sure you will not want to do it on vagrant anymore! :p

I can't use vagrant on my own environment, that's why I try to work around it for now.

hkjn:
> docker build is supported by armbian

Ah, right. That makes sense then.. good luck! :)

hkjn:
> > docker build is supported by armbian
> 
> Ah, right. That makes sense then.. good luck! :)

To add a bit more here, I am very much in favor of add

hkjn:
> > docker build is supported by armbian
> 
> Ah, right. That makes sense then.. good luck! :)

To add a bit more here, I am very much in favor of using the [officially supported](https://docs.armbian.com/Developer-Guide_Building-with-Docker/) Docker tooling to build the Armbian images; I believe it would simplify the experience for new developers to dockerize our toolchain.

After the change to use Docker for the Armbian image build is in, it will be trivial to also dockerize the build process for our Go binaries, at which point any developer with docker installed doesn't need any other requirements to perform the full build.

cc @NicolasDorier: Let us know how your exploration of this topic goes on your end, since we might be able to help out / do research on our own too, depending on other priorities.

PR #29, opened 2019-05-29
Can customize the BOARD

NicolasDorier:
We are using `rock64` so this would allow us to customize the board.

### DIFF
--- a/armbian/build.sh
+++ b/armbian/build.sh
@@ -62,10 +62,11 @@ case ${ACTION} in
 		cp -aR ../../tools userpatches/overlay/					# copy additional software packages to overlay
 		cp -a  ../base/build/customize-image.sh userpatches/	# copy customize script to standard Armbian build hook
 
+		: "${BOARD:=rockpro64}"
 		if [ "${ACTION}" == "build" ]; then
-			vagrant ssh -c 'cd armbian/ && sudo time ./compile.sh BOARD=rockpro64 KERNEL_ONLY=no KERNEL_CONFIGURE=no RELEASE=stretch BRANCH=default BUILD_DESKTOP=no WIREGUARD=no PROGRESS_LOG_TO_FILE=yes LIB_TAG=sunxi-4.20'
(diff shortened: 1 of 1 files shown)


NicolasDorier:
(btw, I gave up with docker, using another powerful server at work)

armbian/build.sh
```diff
@@ -62,10 +62,11 @@ case ${ACTION} in
 		cp -aR ../../tools userpatches/overlay/					# copy additional software packages to overlay
 		cp -a  ../base/build/customize-image.sh userpatches/	# copy customize script to standard Armbian build hook
 
+		: "${BOARD:=rockpro64}"
```
hkjn: Cool, I didn't know about this syntax.. I'd have done:
```
BOARD=${BOARD:-"rockpro64"}
```

Your method produces the same result, but is arguably clearer / less prone to error.. what's this syntax called?

NicolasDorier: That's a very good question lol, I copy pasted from one of my project which I copy pasted from somewhere else when I was learning about bash. :D

Stadicus:
tACK https://github.com/digitalbitbox/bitbox-base/pull/29/commits/1a6d1c3b74070bbdb9e97d271a667c9a235a4258

armbian/build.sh
hkjn: Okay, I found someone who described what the "colon command" is all about:

* https://stackoverflow.com/a/307735

It explains the history of the somewhat weird syntax:

> The Colon Command
>
> I should probably add that the colon command simply has its arguments evaluated and then succeeds. It is the original shell comment notation (before '#' to end of line). For a long time, Bourne shell scripts had a colon as the first character. The C Shell would read a script and use the first character to determine whether it was for the C Shell (a '#' hash) or the Bourne shell (a ':' colon). Then the kernel got in on the act and added support for '#!/path/to/program' and the Bourne shell got '#' comments, and the colon convention went by the wayside. But if you come across a script that starts with a colon, now you will know why.

PR #37, opened 2019-06-04
WIP: Add 'make docker-jekyll' rule

hkjn:
This allows us to render a local version of the
https://base.shiftcrypto.ch/ site.

_Edit:_ There's some unfortunate interactions between Jekyll-in-Docker and VirtualBox-via-Vagrant, see comments.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,3 +1,5 @@
+DOCKER := $(shell which docker 2> /dev/null)
+
 default:
 	mkdir -p build
 	echo "Building tools.."
@@ -6,6 +8,14 @@ default:
 	$(MAKE) -C middleware
 	echo "Building armbian.."
 	$(MAKE) -C armbian
+
+docker-jekyll:
+ifndef DOCKER
+	$(error "This rule requires Docker to run jekyll.")
+endif
+	@echo "Starting docker-jekyll server at localhost:4000.."
+	docker run --rm -it -p 4000:4000 -v $(shell pwd):/srv/jekyll \
+	       jekyll/jekyll:pages jekyll serve --watch --incremental
 
 ci:
 	./scripts/travis-ci.sh
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,5 @@ armbian/base/build/build-local.conf
 tools/bbbfancontrol/bbbfancontrol
 *.sw?
 build/
+_site/
+.jekyll-metadata


Makefile
```diff
@@ -7,5 +7,9 @@ default:
 	echo "Building armbian.."
 	$(MAKE) -C armbian
 
+dev-jekyll:
```
Stadicus: Maybe check if Docker is installed?

Stadicus:
Due to Docker, I now need to run `make` with `sudo`, is that the recommended usage?

If I try to create the local docs with `sudo make dev-jekyll`, I get the following error message:
```
docker: Error response from daemon: pull access denied for ojekyll/jekyll, repository does not exist or may require 'docker login'.
```

hkjn:
> If I try to create the local docs with sudo make dev-jekyll, I get the following error message:

Ah, sorry, a typo snuck in (`ojekyll`, should be `jekyll`). I've fixed it and force-pushed now.

Makefile
hkjn: I've added a check that errors if `which docker` doesn't return a result.

I've also renamed the rule to `docker-jekyll` to more clearly hint that Docker will be required to run the target. I can keep the check as well if you feel it's necessary with the name change, but let me know your preference.

hkjn:
> Due to Docker, I now need to run make with sudo, is that the recommended usage?

That depends on the user's environment (i.e whether they've configured `docker` to be something that requires super-user privileges or not). Most people set things up so their regular user can run `docker` commands (e.g https://github.com/digitalbitbox/bitbox-wallet-app/ used to always assume this), but I prefer to use `sudo` for Docker commands, since an attacker that could launch arbitrary Docker commands could trivially take over the system.

Makefile
Stadicus: That's perfect, thanks!

Stadicus:
> > Due to Docker, I now need to run make with sudo, is that the recommended usage?
> 
> That depends on the user's environment (i.e whether they've configured `docker` to be something that requires super-user privileges or not). Most people set things up so their regular user can run `docker` commands (e.g https://github.com/digitalbitbox/bitbox-wallet-app/ used to always assume this), but I prefer to use `sudo` for Docker commands, since an attacker that could launch arbitrary Docker commands could trivially take over the system.

Yes, I prefer it that way as well.

Stadicus:
Now I get the following error message:
```
$ sudo make docker-jekyll
docker run --rm -it -p 4000:4000 -v /home/stadicus/github/stadicus/bitbox-base:/srv/jekyll \
       jekyll/jekyll:pages jekyll serve --watch --incremental
Unable to find image 'jekyll/jekyll:pages' locally
pages: Pulling from jekyll/jekyll
e7c96db7181b: Pull complete 
622c94c90cb1: Pull complete 
5ab26e9d8a17: Pull complete 
830997f3d72a: Pull complete 
1956a4eaab3f: Pull complete 
d676d5e62937: Pull complete 
Digest: sha256:27847d01b9e11dab8a7c7212eb27928e90bdb0664a31c6d5a596c6d6bef56149
Status: Downloaded newer image for jekyll/jekyll:pages
ruby 2.6.3p62 (2019-04-16 revision 67580) [x86_64-linux-musl]
Configuration file: none
            Source: /srv/jekyll
       Destination: /srv/jekyll/_site
 Incremental build: enabled
      Generating... 
jekyll 3.8.5 | Error:  Permission denied @ rb_sysopen - /srv/jekyll/armbian/armbian-build/ubuntu-bionic-18.04-cloudimg-console.log
Makefile:16: recipe for target 'docker-jekyll' failed
make: *** [docker-jekyll] Error 1
```

hkjn:
> Now I get the following error message:

Right.. what's going on is that the file permissions of some files in the `armbian/armbian-build/` Vagrant are such that Jekyll can't read them, since a Vagrant + VirtualBox VM has created them as super-user inside the VM. I did `cd armbian && make clean` to work around this issue, but it would certainly be preferable to get these tools to cooperate better so that wouldn't be needed.

I tried adding a `.dockerignore` file, but it seems it only takes effect at `docker build` time; when adding volumes, it does not seem to exclude subdirectories from the host..

There seems to be some options to exclude directories for Jekyll, which perhaps would be workable:

* https://jekyllrb.com/docs/configuration/options/

TheCharlatan:
Huh, til: https://github.com/moby/moby/issues/31763

Stadicus:
GitHub Pages is configured to use the `/docs` directory from the `master` branch. Is it possible to just mount this directory inside Docker and run Jekyll with this directory as starting point?

hkjn:
> GitHub Pages is configured to use the /docs directory from the master branch. Is it possible to just mount this directory inside Docker and run Jekyll with this directory as starting point?

I was thinking the same, but it seems that Jekyll expects to be run from the root of the repo:
```
bitbox-base/docs $ sudo docker run --rm -it -p 4000:4000 -v $(pwd):/srv/jekyll jekyll/jekyll:pages jekyll serve --watch --incremental

ruby 2.6.3p62 (2019-04-16 revision 67580) [x86_64-linux-musl]
Configuration file: /srv/jekyll/_config.yml
Invalid theme folder: _includes
Invalid theme folder: _includes
            Source: /srv/jekyll
       Destination: /srv/jekyll/_site
 Incremental build: enabled
      Generating...
Invalid theme folder: _sass
      Remote Theme: Using theme pmarsceill/just-the-docs
   GitHub Metadata: No GitHub API authentication could be found. Some fields may be missing or have incorrect data.
   GitHub Metadata: Error processing value 'title':
  Liquid Exception: No repo name found. Specify using PAGES_REPO_NWO environment variables, 'repository' in your configuration, or set up an 'origin' git remote pointing to your github.com repository. in /_layouts/default.html
             ERROR: YOUR SITE COULD NOT BE BUILT:
                    ------------------------------------
                    No repo name found. Specify using PAGES_REPO_NWO environment variables, 'repository' in your configuration, or set up an 'origin' git remote pointing to your github.com repository.
```

I'll mark this PR as "WIP" for now since there seems to be no obvious way to get Jekyll-in-Docker and VirtualBox-via-Vagrant to play nice with each other.

hkjn:
I'll close this PR for now to reduce clutter.

We can revisit this after switching to Docker for the Armbian OS build which removes the dependency on VirtualBox VMs, see https://github.com/digitalbitbox/bitbox-base/issues/39#issuecomment-499417261 for more.

PR #40, opened 2019-06-05
Add .dockerignore

hkjn:
This allows Docker to avoid copying several GB of data from
`armbian/armbian-build/` which never will be used as build inputs
to the build context, thus speeding up builds.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,1 @@
+armbian/armbian-build/


Stadicus:
utACK https://github.com/digitalbitbox/bitbox-base/pull/40/commits/78dbd4625e8e7fe563dc7402035d72a62a14856f

PR #43, opened 2019-06-05
Makefile: Use the dockerized build for Go tools by default

hkjn:
### Makefile: Use the dockerized build for Go tools by default

Before this change, there was some issue with dependencies
with the default `make` target for the Go tools.

Specifically, in `middleware/main.go` we import other packages:

```
github.com/digitalbitbox/bitbox-base/middleware/src
github.com/digitalbitbox/bitbox-base/middleware/src/handlers
```

The user who has checked out their own fork of the `bitbox-base`
repo might or might not have these packages under the expected path:

```
$GOPATH/src/github.com/digitalbitbox/bitbox-base/middleware/src/{,handlers}
```

A further improvement would be to either `go get` the dependencies we
need, or fail with a better error message, if we can detect this situation
reliably.

Since we have a dockerized build for the Go tools, let's use
that for the default `make` target. I've added a separate
`make build-go` that does the build on the host system (not
using Docker) as well.

### middleware: Also fetch dependencies with 'go get' before build

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,25 +1,31 @@
-.DEFAULT_GOAL=default
+.DEFAULT_GOAL=build-all
 HAS_DOCKER := $(shell which docker 2>/dev/null)
 REPO_ROOT=$(shell dirname $(realpath $(lastword $(MAKEFILE_LIST))))
+
+build-target-exists:
+	@mkdir -p $(REPO_ROOT)/build
 
 check-docker:
 ifndef HAS_DOCKER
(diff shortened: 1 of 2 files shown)


Stadicus:
Nice, this is the first time I was able to just type `make` and out comes a perfect image with everything compiled in. Great work, @hkjn! 

tACK https://github.com/digitalbitbox/bitbox-base/commit/7332141592ba8aa9354af33de8d4fa615fa70319

PR #48, opened 2019-06-06
Improved stability & faster build

Stadicus:
Various minor configuration fixes:
* better, recursive ACL for directory `/mnt/ssd/bitcoin/.bitcoin`
* fix setting hardlinks for bbb-* scripts (now: symlinks)
* remove sha256sum display at end of build script, can take long time with multiple images
* use new [bitbox-base-deps](https://github.com/digitalbitbox/bitbox-base-deps) repository for prebuilt binaries
* by default, c-lightning is compiled during build, but can optionally be installed as package

### DIFF
--- a/armbian/base/build/customize-armbian-rockpro64.sh
+++ b/armbian/base/build/customize-armbian-rockpro64.sh
@@ -46,7 +46,8 @@ CONFIGURATION:
 
 ================================================================================
 BUILD OPTIONS:
-    HDMI OUTPUT:      ${BASE_HDMI_BUILD}
+    BUILD LIGHTNINGD:   ${BASE_BUILD_LIGHTNINGD}
+    HDMI OUTPUT:        ${BASE_HDMI_BUILD}
 ================================================================================
 "
 }
@@ -63,6 +64,7 @@ BASE_SSH_ROOT_LOGIN=${BASE_SSH_ROOT_LOGIN:-"false"}
 BASE_SSH_PASSWORD_LOGIN=${BASE_SSH_PASSWORD_LOGIN:-"false"}
 BASE_DASHBOARD_WEB_ENABLED=${BASE_DASHBOARD_WEB_ENABLED:-"false"}
 BASE_HDMI_BUILD=${BASE_HDMI_BUILD:-"true"}
+BASE_BUILD_LIGHTNINGD=${BASE_BUILD_LIGHTNINGD:-"true"}
 
 # HDMI dashboard only enabled if image is built to support it
 if [[ "${BASE_HDMI_BUILD}" != "true" ]]; then 
@@ -161,11 +163,7 @@ apt update
 apt upgrade -y
 
 # development
-apt install -y  git tmux qrencode bwm-ng
-
-# build c-lightning
-apt install -y  autoconf automake build-essential git libtool libgmp-dev \
-                libsqlite3-dev python python3 net-tools zlib1g-dev libsodium-dev
+apt install -y  git tmux qrencode #bwm-ng
 
 # system
 apt install -y  openssl net-tools fio libnss-mdns \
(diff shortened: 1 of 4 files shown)


armbian/base/build/customize-armbian-rockpro64.sh
```diff
@@ -161,11 +163,7 @@ apt update
 apt upgrade -y
 
 # development
-apt install -y  git tmux qrencode bwm-ng
-
-# build c-lightning
-apt install -y  autoconf automake build-essential git libtool libgmp-dev \
-                libsqlite3-dev python python3 net-tools zlib1g-dev libsodium-dev
+apt install -y  git tmux qrencode #bwm-ng
```
hkjn: Nit: If we don't need the `bmw-ng` package, could we remove the commented-out part of this line?

armbian/base/build/customize-armbian-rockpro64.sh
```diff
@@ -369,15 +367,35 @@ EOF
...
-./configure
-make
-make install
+## either compile c-lightning from source (default), or use prebuilt binary
+if [ "${BASE_BUILD_LIGHTNINGD}" == "true" ]; then
+  apt install -y  autoconf automake build-essential git libtool libgmp-dev \
+                  libsqlite3-dev python python3 net-tools zlib1g-dev libsodium-dev
+
+  cd /usr/local/src/
+  git clone https://github.com/ElementsProject/lightning.git || true
```
hkjn: I know it's not new in this change, but I'm curious if we still need the `|| true` here.. in which circumstances would a failure to clone the repo we're trying to build be okay, and we still would like to continue with the build?

armbian/base/build/customize-armbian-rockpro64.sh
```diff
@@ -369,15 +367,35 @@ EOF
...
+  cd lightning
+  git checkout v${LIGHTNING_VERSION}
+  ./configure
+  make -j 4
+  make install
+
+else
+  apt install -y libsodium-dev
```
hkjn: (Optional) Since we install `libsodium-dev` in both branches of this `if` statement, it's not really conditional, so we could also install it along with other dependencies higher up in the script and remove it from both branches.

armbian/base/build/customize-armbian-rockpro64.sh
```diff
@@ -369,15 +367,35 @@ EOF
...
+  cd /usr/local/src/
+  # temporary storage of 'lightningd' until official arm64 binaries work with stable Armbian release
+  curl --retry 5 -SLO https://github.com/digitalbitbox/bitbox-base-deps/releases/download/${BIN_DEPS_TAG}/lightningd_${LIGHTNING_VERSION}-1_arm64.deb
+  if ! echo "52be094f8162749acb207bf9ad08125d25288a9d03eb25690f364ba42fcff3c4  lightningd_0.7.0-1_arm64.deb" | sha256sum -c -; then
+    echo "sha256sum for precompiled 'lightningd' failed" >&2
+    exit 1
+  fi
+  dpkg -i lightningd_${LIGHTNING_VERSION}-1_arm64.deb
+  ln -s /usr/bin/lightningd /usr/local/bin/lightningd
```
hkjn: I'm curious why this symlink is needed.. is the `make install` going to put `lightningd` in a different path than installing from the `.deb` package?

armbian/base/build/customize-armbian-rockpro64.sh
Stadicus: I'd like to keep it in the c-lightning section, but can move it in front of the if-clause.

armbian/base/build/customize-armbian-rockpro64.sh
Stadicus: Yes, exactly. I'll add a comment. The `make install` command puts the binaray in `/usr/local/bin`, while the deb package installs into `/usr/bin`...

armbian/base/build/customize-armbian-rockpro64.sh
hkjn: SGTM.

armbian/base/build/customize-armbian-rockpro64.sh
hkjn: Cool, adding a comment sounds useful then.

armbian/base/build/customize-armbian-rockpro64.sh
```diff
@@ -479,7 +501,8 @@ cp /tmp/overlay/bbbfancontrol /usr/local/sbin/
 cp /tmp/overlay/bbbfancontrol.service /etc/systemd/system/
 
 ## base-middleware
-cp /tmp/overlay/base-middleware /usr/local/sbin/
+## see https://github.com/digitalbitbox/bitbox-base/blob/master/middleware/README.md
+# cp /tmp/overlay/base-middleware /usr/local/sbin/
```
hkjn: Looks like we accidentally commented out this line. Sending PR to fix..

hkjn: PR that restores the line: https://github.com/digitalbitbox/bitbox-base/pull/51

PR #53, opened 2019-06-11
Add docs for bitcoind / c-lightning / electrs

Stadicus:


docs/applications/bitcoin-core.md
```diff
@@ -0,0 +1,125 @@
+---
+layout: default
+title: Bitcoin Core
+parent: Main Applications
+nav_order: 610
+---
+## Bitcoin Core
+
+Bitcoin Core the most popular software implementation of Bitcoin, openly developed on GitHub (<https://github.com/bitcoin/bitcoin>) and publishing releases on <https://bitcoincore.org>. The BitBox base uses the latest stable binary release to communicate with the Bitcoin peer-to-peer network, for example to learn about current transactions and receive newly mined blocks. All data is regarded as untrusted until it is locally verified to the consensus rules enforced by the BitBox Base.
```
hkjn: I feel that saying that `the consensus rules enforced by the BitBox Base` in the context of describing Bitcoin Core is a little odd.. why not say that `the consensus rules verified by Bitcoin Core`?

(I.e `Bitcoin Core`, not `BitBox Base`, as it is the component that verifies the rules, and `verified`, not `enforced`, as there's no real way to force the rest of the Bitcoin network to do anything by running a certain program, but we can verify that the blocks and transactions we receive are following the correct rules.)

hkjn: Nit: s/`BitBox base`/`BitBox Base`

docs/applications/bitcoin-core.md
```diff
@@ -0,0 +1,125 @@
...
+
+### Using binary releases
+
+While the `master` branch on GitHub reflects the current development process, a new stable release is tagged and tested every few months. This release is built from the source code in a reproduciable way for various platforms by many contributors, verifying that they get the exact same binary executables by comparing the sha256 hash. These hash values for all images are provided in the `SHA256SUM.asc` file, which is signed by a Bitcoin Core maintainer, currently [laanwj](https://github.com/laanwj).
```
hkjn: Typo: s/`reproduciable`/`reproducible`

docs/applications/bitcoin-core.md
```diff
@@ -0,0 +1,125 @@
...
+
+Instead of compiling Bitcoin Core from source when building the Armbian image, or poviding our own binaries, we believe that following the official binary releases is a good way to minimize trusting any single entity and keep the build process performant.
+
+To ensure that the official binary is used, we store the verified signing key `90C8 019E 36C2 E964` independently from the Bitcoin Core release site and validate the downloaded binary against the signed hash values. If this verification does not succeed (a single bit difference would be enough) the build script abort with an error.
```
hkjn: Nit: s/`abort`/`aborts`

docs/applications/bitcoin-core.md
```diff
@@ -0,0 +1,125 @@
...
+
+### Configuration
+
+The application configuration is specified in the local `/etc/bitcoin/bitcoin.conf` file. Please check the most current initial configuration in [`customize-armbian-rockpro64.sh`](../../armbian/base/build/customize-armbian-rockpro64.sh).
```
hkjn: The second sentence is a little cumbersome (particularly, `the most current initial configuration`), and might be more clear if reworded as:

> The `bitcoin.conf` file is written into the Armbian image by the [`customize-armbian-rockpro64.sh`](../../armbian/base/build/customize-armbian-rockpro64.sh) script.

docs/applications/bitcoin-core.md
```diff
@@ -0,0 +1,125 @@
...
+
+```
```
hkjn: I'm curious if we think there might be a risk of maintaining a copy of the `bitcoin.conf` inline in this doc as well as inside the `customize-armbian-rockpro64.sh` script. Alternatively, perhaps we could check in the file separately, and read it from `customize-armbian-rockpro64.sh` as well as either linking to it here (or even inlining it using some preprocessor script to generate the `.md`).

(And elsewhere.)

docs/applications/bitcoin-core.md
```diff
@@ -0,0 +1,125 @@
...
+# tor
+proxy=127.0.0.1:9050
+seednode=nkf5e6b7pl4jfd4a.onion
+seednode=xqzfakpeuvrobvpj.onion
+seednode=tsyvzsqwa2kkf6b2.onion
+```
+
+Some notes about this specific configuration:
+
+* **Application options**
```
hkjn: Similar to my comment above, ideally we'd find a way to put this (excellent!) documentation around the different `bitcoin.conf` options very near the actual config. Perhaps it would be possible to add this information directly inline in the `bitcoin.conf` file, and then inline it inside this `.md` file? We could do this in a follow-up PR, but wanted to raise the idea.

docs/applications/bitcoin-core.md
Stadicus: changed to `...until it is locally verified to match the consensus rules of Bitcoin Core`

docs/applications/bitcoin-core.md
Stadicus: Yes, that would be interesting. Over time, it might become outdated here, but let's discuss this as one of the next steps.

docs/applications/bitcoin-core.md
Stadicus: Agreed.

docs/applications/bitcoin-core.md
```diff
@@ -0,0 +1,125 @@
...
+
+### Using binary releases
+
+While the `master` branch on GitHub reflects the current development process, a new stable release is tagged and tested every few months. This release is built from the source code in a reproducible way for various platforms by many contributors, verifying that they get the exact same binary executables by comparing the sha256 hash. These hash values for all images are provided in the `SHA256SUM.asc` file, which is signed by a Bitcoin Core maintainer, currently [laanwj](https://github.com/laanwj).
+
+Instead of compiling Bitcoin Core from source when building the Armbian image, or poviding our own binaries, we believe that following the official binary releases is a good way to minimize trusting any single entity and keep the build process performant.
+
+To ensure that the official binary is used, we store the verified signing key `90C8 019E 36C2 E964` independently from the Bitcoin Core release site and validate the downloaded binary against the signed hash values. If this verification does not succeed (a single bit difference would be enough) the build script aborts with an error.
```
hkjn: (Optional) Might be good to list the full PGP fingerprint ("long id" of 8 bytes is much better than "short id" of 4 bytes, but "fingerprint" of 20 bytes is usually seen as best practice):

* `01EA 5486 DE18 A882 D4C2 6845 90C8 019E 36C2 E964`

Stadicus: done!

Issue #39, opened 2019-06-05
Switch to Docker for Armbian build


hkjn:
In our current build process, we use VirtualBox VMs managed by Vagrant to build the Armbian-based OS images for the Base:

* https://base.shiftcrypto.ch/os/armbian-build.html

This has some downsides:

* all developers or automated systems need Vagrant and VirtualBox installed, with specific versions; in comparison, Docker is becoming a standard dependency to manage dependencies of various tech stacks for many developers
* we already use Docker in the Base project for [building Go programs](https://github.com/digitalbitbox/bitbox-base/blob/master/Dockerfile), so using it for Armbian builds would reduce the set of dependencies
* having several virtualization technologies in the same repo has already led to [unfortunate interactions / extra complexity](https://github.com/digitalbitbox/bitbox-base/pull/37#issuecomment-498755381), so just relying on Docker would be preferable

This work was earlier discussed in a [PR comment thread](https://github.com/digitalbitbox/bitbox-base/pull/26#issuecomment-496948578).

The official Armbian docs for building with Docker are available at:

* https://docs.armbian.com/Developer-Guide_Building-with-Docker/

This issue can be closed once support for building Armbian OS images with Docker has been merged to `master`, or if we decide not to pursue this.

hkjn:
I tried making the changes to `armbian/build.sh` to switch to Docker-based builds for Armbian OS images in https://github.com/digitalbitbox/bitbox-base/commit/6fbda4cd15297949c134c9060f5e05061a2eae59.

The [branch](https://github.com/digitalbitbox/bitbox-base/compare/master...hkjn:20190605-dockerize-armbian?expand=1) seems to build fine locally, and produces an `.img` file that boots on a RockPro64 board and which seems functional on cursory examination, however on Travis the build errors:

* https://travis-ci.org/hkjn/bitbox-base/builds/542162183

The symptom seen on Travis builds is: 
```
# Successfully built b1561d8cbb3b
# Successfully tagged armbian:latest
# [ o.k. ] Running the container 
# docker: Error response from daemon: create ./: "./" includes invalid characters for a local volume name, only "[a-zA-Z0-9][a-zA-Z0-9_.-]" are allowed. If you intended to pass a host directory, use absolute path.
# See 'docker run --help'.
# [ error ] Docker command failed, check syntax or version support. Error code:   [ 125 ]
```

[...]

PR #56, opened 2019-06-13
Dockerize armbian build

hkjn:
Note that we do not switch over Travis CI to use the dockerized build yet, since some issues remain:

* https://github.com/digitalbitbox/bitbox-base/issues/39#issuecomment-501343881

After this change, the default `make` command in the root of the repo (`make build-all`) uses Docker to build the Armbian image as well. (We already were building the Go services using Docker since ac5a3040.)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,6 +21,10 @@ build-all: build-target-exists docker-build-go
 	@echo "Building armbian.."
 	$(MAKE) -C armbian
 
+clean:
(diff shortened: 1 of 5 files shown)


Stadicus:
Great work @hkjn! tACK https://github.com/digitalbitbox/bitbox-base/pull/56/commits/dedde7939c0453f47dd33b3c6854682935d451dc

PR #58, opened 2019-06-13
docs: Add build-details.md doc

hkjn:
This piece of documentation goes into a bit more depth about
how the Armbian build process works, specifically which scripts
copy which files from host -> overlay -> Armbian filesystem,
and which scripts have which responsibilities.

~_Edit:_ This PR doesn't directly depend on https://github.com/digitalbitbox/bitbox-base/pull/56, but it does describe the Armbian build process as being dockerized, so probably would make sense to merge that PR first.~

### DIFF
--- a/docs/os/build-details.md
+++ b/docs/os/build-details.md
@@ -1,0 +1,30 @@
+---
+layout: default
+title: Build details
+parent: Operating System
+nav_order: 315
+---
+## Armbian build details
+
+This document goes into more detail on how the Armbian build process works.
+
+The [`armbian/build.sh`](https://github.com/digitalbitbox/bitbox-base/blob/master/armbian/build.sh) script is the entrypoint to the Armbian build process. When running `make` in the `armbian/` directory, it in turn invokes the `armbian/build.sh` script under the hood.
+
+It performs the following steps:
+
+1. cloning the [`github.com/armbian/build`](https://github.com/armbian/build/) repo into `armbian/armbian-build/`, if it doesn't exist already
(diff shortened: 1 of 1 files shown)


docs/os/build-details.md
```diff
@@ -0,0 +1,30 @@
+---
+layout: default
+title: Build details
+parent: Operating System
+nav_order: 210
```
Stadicus: nav_order should be between 310 (Building the Armbian base image) and 320 (Configuration), so maybe 315?

docs/os/build-details.md
```diff
@@ -0,0 +1,30 @@
...
+---
+## Armbian build details
+
+This document goes into more detail on how the Armbian build process works.
+
+The [`armbian/build.sh`](https://github.com/digitalbitbox/bitbox-base/blob/master/armbian/build.sh) script is the entrypoint to the Armbian build process.
```
Stadicus: To have a link from the previous chapter (Building the Armbian base image) that only references the `make` command, we could note that `make` calls the `build.sh` script.

docs/os/build-details.md
hkjn: Done.

docs/os/build-details.md
hkjn: Sure, added.

hkjn:
Thanks for the review, @Stadicus!

I've force-pushed https://github.com/digitalbitbox/bitbox-base/commit/7b1a3ad5dab6bcafeff624428fb952e9459fe818 which addresses your review comments, PTAL.

Stadicus:
ACK

PR #68, opened 2019-06-14
docs: trivial s/TOR/Tor change

hkjn:
(It seems silly, but the capitalization of the the project's name
for whatever reason does seem to be a shibboleth / thing that rubs
people the wrong way.)


Stadicus:
ACK

PR #69, opened 2019-06-14
Fix some broken links, add Go applications to index.md

hkjn:
As an aside, we might want to set up some quick scraper as part of CI to check that there's no links that 404s, since checking manually might become cumbersome..

### DIFF
--- a/docs/os/build-details.md
+++ b/docs/os/build-details.md
@@ -15,7 +15,7 @@ It performs the following steps:
 1. cloning the [`github.com/armbian/build`](https://github.com/armbian/build/) repo into `armbian/armbian-build/`, if it doesn't exist already
 1. copying over the following files from the host system into `armbian/armbian-build/userpatches/overlay/`:
     1. `armbian/base/`: scripts and configs included in the Armbian image
-    1. `build/`: contains all [built Go binaries](/docs/go-apps.md) and their associated `.service` files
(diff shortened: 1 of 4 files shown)


docs/os/build-details.md
```diff
@@ -15,7 +15,7 @@ It performs the following steps:
 1. cloning the [`github.com/armbian/build`](https://github.com/armbian/build/) repo into `armbian/armbian-build/`, if it doesn't exist already
 1. copying over the following files from the host system into `armbian/armbian-build/userpatches/overlay/`:
     1. `armbian/base/`: scripts and configs included in the Armbian image
-    1. `build/`: contains all [built Go binaries](/docs/go-apps.md) and their associated `.service` files
+    1. `build/`: contains all [built Go binaries](go-apps.md) and their associated `.service` files
```
Stadicus: Not sure that works out of the `/docs/os/` directory?

hkjn: Yeah, I think you're right. On the live site it seems that the URL is https://base.shiftcrypto.ch/go, so I've updated this accordingly. It seems to render as `http://localhost:4000/go` on local Jekyll, so I think the new updated link should work.

hkjn:
Thanks for the review @Stadicus, PTAL.

Stadicus:
utACK https://github.com/digitalbitbox/bitbox-base/commit/03f090a6bf1ea1cc0c64f0c59c7d6c1bcfdd80cf

PR #77, opened 2019-06-17
bitcoind: signing key & Tor listening

Stadicus:
* get Bitcoin Core signing key from keyserver instead of downloading it directly
* make `bitcoind` listen over Tor directly

Built & tested on Alice

### DIFF
--- a/armbian/base/build/customize-armbian-rockpro64.sh
+++ b/armbian/base/build/customize-armbian-rockpro64.sh
@@ -268,7 +268,14 @@ fi
 apt update
 apt -y install tor --no-install-recommends
 
+# allow user 'bitcoin' to access Tor proxy socket
+usermod -a -G debian-tor bitcoin
+
 cat << EOF > /etc/tor/torrc
+ControlPort 9051                                          #TOR#
+CookieAuthentication 1                                    #TOR#
+CookieAuthFileGroupReadable 1                             #TOR#
+
 HiddenServiceDir /var/lib/tor/hidden_service_bitcoind/    #BITCOIND#
 HiddenServiceVersion 3                                    #BITCOIND#
 HiddenServicePort 18333 127.0.0.1:18333                   #BITCOIND#
@@ -317,6 +324,7 @@ testnet=1
 # server
 server=1
 listen=1
+listenonion=1
 daemon=1
 txindex=0
 prune=0
--- a/armbian/base/build/customize-armbian-rockpro64.sh
+++ b/armbian/base/build/customize-armbian-rockpro64.sh
@@ -168,7 +168,7 @@ apt install -y  git tmux qrencode
 # system
 apt install -y  openssl net-tools fio libnss-mdns \
                 avahi-daemon avahi-discover avahi-utils \
-                fail2ban acl ifmetric 
+                fail2ban acl ifmetric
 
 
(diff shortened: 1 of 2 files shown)


Stadicus:
Now on `bitcoind` startup, a hidden service is created:
```
2019-06-17T13:50:50Z tor: Got service ID **********, advertising service **********.onion:18333
2019-06-17T13:50:50Z AddLocal(**********.onion:18333,4)
```

armbian/base/build/customize-armbian-rockpro64.sh
```diff
@@ -299,10 +306,12 @@ BITCOIN_VERSION="0.18.0"
 
 mkdir -p /usr/local/src/bitcoin
 cd /usr/local/src/bitcoin/
-curl --retry 5 -SL https://bitcoincore.org/keys/laanwj-releases.asc | gpg --import
+#curl --retry 5 -SL https://bitcoincore.org/keys/laanwj-releases.asc | gpg --import
```
hkjn: Nit: Could we remove commented-out lines?

armbian/base/build/customize-armbian-rockpro64.sh
```diff
@@ -299,10 +306,12 @@ BITCOIN_VERSION="0.18.0"
...
 curl --retry 5 -SLO https://bitcoincore.org/bin/bitcoin-core-${BITCOIN_VERSION}/SHA256SUMS.asc
 curl --retry 5 -SLO https://bitcoincore.org/bin/bitcoin-core-${BITCOIN_VERSION}/bitcoin-${BITCOIN_VERSION}-aarch64-linux-gnu.tar.gz
 
+## get Bitcoin Core signing key
+gpg --keyserver pool.sks-keyservers.net --recv-keys 01EA5486DE18A882D4C2684590C8019E36C2E964
```
hkjn: (Optional) Another option would be to just add the full PGP key to this repository, thus avoiding any network calls. (GPG keyservers can be notoriously unstable.)

docs/support/prometheus.md
```diff
@@ -0,0 +1,112 @@
+---
```
hkjn: I believe this file was added to `master` in 870d93e28f, so it's surprising to see it in this PR.. perhaps the branch could be rebased on the latest `master`?

armbian/base/build/customize-armbian-rockpro64.sh
Stadicus: ACK

armbian/base/build/customize-armbian-rockpro64.sh
Stadicus: I think receiving the key from public server makes it more transparent & trustless (no need to check the actual keyfile). But I can live with both solutions.

hkjn: Okay, let's keep it like this for now.

armbian/base/build/customize-armbian-rockpro64.sh
```diff
@@ -317,6 +325,7 @@ testnet=1
 # server
 server=1
 listen=1
+listenonion=1
```
hkjn: (Optional) Not necessarily for this PR, but perhaps we should also update the docs accordingly, since we have an annotated `bitcoin.conf` there?

PR #78, opened 2019-06-18
Add laanwj-releases.asc file

Stadicus:
Using keyservers to receive the Bitcoin signing key is very unreliable so I include the full public keyfiles in the build process for importing. Keys are still updated using keyservers as the next step, but the script will continue even if that fails. Worst case: buidl breaks due to outdated signing key.

Additional minor change:
* add additional error messages to systemd-startup-scripts.sh

Build tested successfully.


Stadicus:
Did some `git` magic and the file should now be present.

PR #79, opened 2019-06-18
Simplify build rules

hkjn:
### Makefile: add build/README.md, simplify build rules

By adding a build/README.md file, we can guarantee that the build/
directory always will exist, so we no longer need the
`build-target-exists` make targets.

h/t @NicolasDorier, who had questions prompting us to rethink this
setup which led to this patch.

### armbian: Remove trivial build.sh clean command

Instead just inline the command in the `Makefile`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,7 +1,6 @@
 armbian/armbian-build
 armbian/base/build/build-local.conf
 armbian/base/config/latest_commit
-build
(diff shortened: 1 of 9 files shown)


Stadicus:
utACK https://github.com/digitalbitbox/bitbox-base/pull/79/commits/b37f2fbabda19b3cace7bca57bd4b9bd1e8e8869

PR #81, opened 2019-06-18
tools/bbbsupervisor: Rework supervisor to avoid 3rd party deps

hkjn:
With this change, we have no 3rd party dependencies remaining for
bbbsupervisor.go.

If we fail to start the `journalctl` command or receive any output
on stderr, we currently pass on the message and crash with non-success
exit status.

Also enable building bbbsupervisor in tools/Makefile.

### DIFF
--- a/tools/bbbsupervisor/bbbsupervisor.go
+++ b/tools/bbbsupervisor/bbbsupervisor.go
@@ -41,7 +41,6 @@
 // * NGINX, Grafana, ...
 //   - monitor service availability
 //
-//
 
 package main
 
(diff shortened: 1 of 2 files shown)


tools/bbbsupervisor/bbbsupervisor.go
```diff
@@ -102,38 +112,41 @@ func main() {
 	version := flag.Bool("version", false, "return program version")
 	flag.Parse()
 
-	fmt.Println("bbbsupervisor version", versionNum)
+	fmt.Printf("bbbsupervisor version %v starting..\n", versionNum)
```
Stadicus: The idea of this version printout was to have it on start (in the log or stdout) and also use it for `bbbsupervisor --version` (thus the Exit 0) immediately after. With "starting..." this no longer is phrased right. I'd prefer to keep it as before.

hkjn: Ah, I didn't think about the behavior when called with `-version`. Good point, fixed.

hkjn:
Thanks for the review, @Stadicus.

I force-pushed d4e0f8d5ae0e1816e4fcfb0b42efb4cc5256c1e6, PTAL.

PR #82, opened 2019-06-18
armbian/base: Allow ICMP/ping

hkjn:
I'm not sure why these ended up as commented out in aa44547af9679cfbdf7b945fc2093eeef2992da6, but it wasn't intentional, and the current state doesn't respond to `ping` since the packets end up being `DROP`'ed.


Stadicus:
utACK https://github.com/digitalbitbox/bitbox-base/pull/82/commits/98771a3dd4d2772926955fb77728223f468fcd90

PR #84, opened 2019-06-19
Improve docker build and caching

NicolasDorier:
@hkjn,

This PR achieve the following:
* Cleanup the docker image during `make clean`
* Make efficient use of docker cache system by building the tools inside the `Dockerfile` and the use of multi stage build
* `docker-build-go` is only building (no build if docker cache hit) and copying the build output to the host's mapped folder

This make sure that modifications only to tooling for example does not rebuild the middleware.
If no file changes then `docker-build-go` is near instantaneous.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,28 +2,36 @@
 # The docker image produced by this config provides the build environment
 # for BitBox Base.
 #
-FROM ubuntu:18.04
-
+FROM golang:1.12.5-stretch as bitbox-base
 RUN apt-get update && apt-get install -y --no-install-recommends \
     clang \
     gcc \
     libc6-dev \
     make \
-    curl \
-    ca-certificates \
-    git
+    ca-certificates 
+WORKDIR /go/src/github.com/digitalbitbox/bitbox-base
+RUN mkdir build
 
-RUN mkdir -p /opt/go_dist &&\
-    curl -O https://dl.google.com/go/go1.12.5.linux-amd64.tar.gz &&\
-    echo "aea86e3c73495f205929cfebba0d63f1382c8ac59be081b6351681415f4063cf go1.12.5.linux-amd64.tar.gz" | sha256sum -c &&\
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -12,18 +11,29 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
     curl \
     ca-certificates \
     git
```
hkjn: (Optional) Seems like it would longer be necessary to install `git` and `curl` packages now, as they are installed in the `golang:1.12.5-stretch` image already:

```
sudo docker run --rm -it golang:1.12.5-stretch bash -c 'which curl && which
git'
/usr/bin/curl
/usr/bin/git
```

NicolasDorier:
The `docker rmi` does not clean the cache, seems I need `docker rmi $(docker images -a --filter=dangling=true -q)` But this will delete the cache of all images on the system :(

hkjn:
> The `docker rmi` does not clean the cache, seems I need `docker rmi $(docker images -a --filter=dangling=true -q)` But this will delete the cache of all images on the system :(

Yeah, I think between those two options it's better to just remove the final image and not the cached layers used to produce it for `make clean`.. if there are issues that requires cached layers to be removed as well, users could in that case manually take steps wipe out all their untagged images / all the Docker data. 

(Optional) Possibly this limitation could be mentioned as a `TODO` in the `make clean` rule?

NicolasDorier:
Added a comment

Dockerfile
NicolasDorier: Removed curl and git

NicolasDorier:
@hkjn btw, the cool stuff of this Dockerfile also, is that it is multi arch. You can build it on arm just fine as well!

hkjn:
> @hkjn btw, the cool stuff of this Dockerfile also, is that it is multi arch. You can build it on arm just fine as well!

Good point, and I can confirm; I just tested `golang:1.12.5-stretch` on `aarch64` as well as `armv7l`, and it worked on both.

PR #85, opened 2019-06-19
docs: Make docs.armbian.com a link

hkjn:


hkjn:
> Just wrapping the link in `<http://www.example.com>` should work as well.

Thanks, that's certainly simpler! Fixed.

PR #87, opened 2019-06-20
Makefile: Add docker-jekyll rule

hkjn:
Second attempt to make it easier to serve https://base.shiftcrypto.ch/ locally during development. (See https://github.com/digitalbitbox/bitbox-base/pull/37 for the first one.)

We no longer use Vagrant for builds, so the issues seen in #37 no longer should affect us.

We have another odd interaction however between the dockerized Armbian build process and Jekyll, where Jekyll appears to try to read a non-existing directory. In this change, we work around that issue by removing the parent directory from the build cache, and we also add a TODO to describe how this could be improved.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,7 +1,16 @@
+# Ignore the github.com/armbian/build repo we check out during builds.
 /armbian/armbian-build
+# Ignore the local configuration overrides for Armbian build.
 /armbian/base/build/build-local.conf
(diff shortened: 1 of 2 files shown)


Stadicus:
Thanks for adding the comments and todo.

utACK https://github.com/digitalbitbox/bitbox-base/pull/87/commits/d7b69862702eabca9ddf350384a0d932dbad804e

PR #90, opened 2019-06-24
Add Python checks to CI, fix all style issues

hkjn:
This PR addresses all style issues reported by the `black` and `pylint` tools with out-of-the-box configs*, and enabled those checks on CI (travis-ci.org). We also move the `Dockerfile` into `scripts/`, and add a new `Dockerfile-python-ci`, since some of these tools were cumbersome to cram into the main image.

* Except for the line length, which we override to 120 colums.


hkjn:
Closing as defunct (presumably, had forgotten this was open).

PR #91, opened 2019-06-24
tools/bbbsupervisor: Implement basic prometheusFollower

hkjn:
Like `logFollower`, a `prometheusFollower` has a basic `follow()`
method, and when called to follow, periodically queries a Prometheus
server for a particular expression.


hkjn:
See also #91 for an alternative PR, which does not have the 3rd party dependencies this one does.

hkjn:
Closing in favor of #91 after discussing offline with @Stadicus.

We believe that the maintenance cost of using the 3rd party dependency (`github.com/prometheus/client_golang` and friends) outweighs the benefit, and just redefining the JSON structure we receive over the HTTP API as in #91 should be easier to work with in the future.

PR #94, opened 2019-06-25
.dockerignore: Also ignore _site/

hkjn:
This directory is present if the user has run `make docker-jekyll`,
added in 5b3477c4, and can be multiple GB. Excluding it from Docker
builds speeds things up significantly.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,1 +1,2 @@
 armbian/armbian-build/
+_site/


Stadicus:
utACK https://github.com/digitalbitbox/bitbox-base/pull/94/commits/c581fb49412706af212ae8fdf152f16fa4d619b2

PR #95, opened 2019-06-25
Downsize Armbian disk image to ~1 GB

Stadicus:
Previously, the Armbian disk image was > 2 GB. Removing many unnecessary packages allowed to shrink the Armbian image by 50% to ~1 GB. As Mender uses compressed update images, the data to download for a full disk image update will be < 300 MB.

Includes minor improvements that popped up during testing:
* electrs: start binary directly from electrs.service
* c-lightning: add slight delay before setting group permissions on rpc

### DIFF
--- a/armbian/base/build/customize-armbian-rockpro64.sh
+++ b/armbian/base/build/customize-armbian-rockpro64.sh
@@ -476,8 +476,17 @@ Description=Electrs server daemon
 Wants=bitcoind.service
 After=bitcoind.service
 [Service]
+EnvironmentFile=/etc/electrs/electrs.conf
+EnvironmentFile=/mnt/ssd/bitcoin/.bitcoin/.cookie.env
 ExecStartPre=/bin/systemctl is-active bitcoind.service
-ExecStart=/opt/shift/scripts/systemd-electrs-start.sh
+ExecStart=/usr/bin/electrs \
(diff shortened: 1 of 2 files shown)


hkjn:
@Stadicus: Is this PR ready for review? (I'm asking because the 9ad9c3217c999dff139ae346b481f62829f695c2 commit mentions `(wip)` in the title.)

Stadicus:
Good point. Yes, it's ready for review.

armbian/base/build/customize-armbian-rockpro64.sh
```diff
@@ -162,11 +162,7 @@ apt purge -y ntp network-manager
 apt update
 apt upgrade -y
 
-# development
-apt install -y  git tmux qrencode
-
-# system
-apt install -y  openssl net-tools fio libnss-mdns \
+apt install -y  git openssl net-tools fio libnss-mdns \
```
hkjn: (Optional) Not new in this PR, but I noticed we don't specify `--no-install-recommends` here, as we do elsewhere. Would it make sense to add it here, to only install packages we explicitly list?

Stadicus: Good point, I'll tackle that in a next pull request with proper testing.

PR #98, opened 2019-06-25
Use host user for Docker builds

hkjn:
By using the same user inside the container as on the host, we avoid
issues with the host user lacking the privileges to e.g remove files
under `build/`, which are bind-mounted from the host into the container.

Before this change, those files would run as the super-user (`id -u` of
`0`), which caused the issues mentioned above, specifically on Gitlab,
where subsequent test runners would find themselves unable to clean up
the `build/` artifacts from prior builds.

The unfortunate edge case here is for users that (wisely!) require
`sudo` to run `docker` commands, there is no way for the `Makefile`
to deduce what their non-privileged user's id is. Those users
need to either deal with the super-user (id `0`, commonly `root`)
owning the files under `build/`, or to explicitly specify their
non-privileged user's id when calling `make` commands:

```
sudo make BUILDER_UID=$(id -u)
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -32,6 +32,8 @@ RUN make -C "tools"
 
 # Final
 FROM golang:1.12.5-stretch as final
-
+ARG builder_uid
+RUN adduser --disabled-password --gecos "" --uid ${builder_uid} builder
+USER builder
 COPY --from=middleware-builder /go/src/github.com/digitalbitbox/bitbox-base/build/. /opt/build/.
 COPY --from=middleware-tools /go/src/github.com/digitalbitbox/bitbox-base/build/. /opt/build/.
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,7 @@
 .DEFAULT_GOAL=build-all
 HAS_DOCKER := $(shell which docker 2>/dev/null)
 REPO_ROOT=$(shell dirname $(realpath $(lastword $(MAKEFILE_LIST))))
+BUILDER_UID=$(shell id -u)
 
 check-docker:
 ifndef HAS_DOCKER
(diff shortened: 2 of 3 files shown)


hkjn:
This PR can be tested by removing the `build/` directory entirely, then restoring the `build/README.md` from the repo (i.e bring the repo back to a clean state), then doing a build, and observing that the files are owned by the non-privileged user:

```
$ sudo rm -rf build/
$ git checkout -- build/README.md
$ sudo make docker-build-go BUILDER_UID=$(id -u)
$ ls -hsal build/
total 14M
4.0K drwxr-xr-x  2 user user 4.0K Jun 25 17:45 .
4.0K drwxr-xr-x 13 user user 4.0K Jun 25 17:50 ..
8.6M -rwxr-xr-x  1 user user 8.6M Jun 25 17:45 base-middleware
2.3M -rwxr-xr-x  1 user user 2.3M Jun 25 17:45 bbbfancontrol
4.0K -rw-r--r--  1 user user  227 Jun 25 17:45 bbbfancontrol.service
2.5M -rwxr-xr-x  1 user user 2.5M Jun 25 17:45 bbbsupervisor
4.0K -rw-r--r--  1 user user  194 Jun 25 17:45 bbbsupervisor.service
4.0K -rw-r--r--  1 user user  167 Jun 25 17:45 README.md
``

**Comment by Stadicus** [comment]
utACK https://github.com/digitalbitbox/bitbox-base/commit/736aa3defcf6d9dd4250f5b9ef2bb6700c75abae

File: Dockerfile [hunk 1]
```diff
@@ -32,6 +32,8 @@ RUN make -C "tools"
 
 # Final
 FROM golang:1.12.5-stretch as final
-
+ARG builder_uid
+RUN adduser --disabled-password --gecos "" --uid ${builder_uid} builder
```
hkjn: This commit actually introduced a regression, where if users have `sudo` required to run `docker` commands, their build will fail on this line. I've sent https://github.com/digitalbitbox/bitbox-base/pull/99 to revert this commit for now, as we figure out the best path forward.

PR #105, opened 2019-07-01
General stability improvements and scripting cleanup

Stadicus:
Minor improvements
* disable UART serical console output, as a preparation for HSM connectivity
* disable Armbian ZRAM swapfile and replace it with custom swapfile on SSD (includes additional Grafana charts)
* allow `bbb-config.sh exec bitcoind_reindex` to wipe chainstate and reindex downloaded blocks
* use standard `network-manager` instead of hardcoded files
* remove unnecessary Avahi service announcements
* general cleanup of customize script


Stadicus:
Built & tested https://github.com/digitalbitbox/bitbox-base/pull/105/commits/2bb4994204e3a6048b2568c4dbdad97de736d1d9

PR #117, opened 2019-07-24
General stability and reliability improvements

Stadicus:
**Read-only filesystem improvements**
This pull request builds on the raw read-only root filesystem configuration. The following improvements / features are added, while image creation without `overlayroot` is preserved.
* `overlayroot` is only used when build option is active and Ubuntu is used
* disable Armbian ramlog as /var/log is part of the ramdisk anyway
* add TMPF measure to monitor the ramdisk with Prometheus

User configuration uses the `/data` directory. Build configuration is stored in the `/data_source` directory. On regular images, a symlink `/data --> /data_source` to created on first boot. On a Mender-enabled image, however, `/data` is a separate writable partition preserved over updated. Build configuration is copied there on first boot.
* store the hostname in `/data/network/hostname`
* create selfsigned ssl certificates & keys in `/data/ssl`
* create `/data/trigger/` directory for file triggers used by systemd
* move sysconfig directory to `/data/sysconfig`

**General optimization**
* bitcoind dbcache is set to 300MB and needs to be increased by bbbsupervisor (todo)
* c-lightning and electrs are not started until bitcoind IBD is finished
* systemd journal is configured for persistant log storage on SSD
* logrotation is sticter: file sizes are limited to 5M and checked every 10 min
* reduce logging of NGINX (stricter log rotation and no access logs)
* add stresstest-ssd.sh script
* u-boot config disables the serial console ttyS2 over UART and uses only the display (prerequisite to use the BitBox secure platform over this UART port)
* prometheus-bitcoind.py logs only ever 5 min during IBD, every 30 secs after that
* remove bitcoind mempool limitation to avoid skewing fee calculation and statistics
* add `PartOf=bitcoind.service` to lightningd and electrs units to automatically restart them when bitcoind restarts

**Applications**
* Prometheus and Node Exporter are updated to fix an issue that two hours of metrics are lost on reboot, see https://github.com/prometheus/tsdb/issues/590
* bbb-config.sh: add command `set bitcoin_dbcache`
* bbb-config.sh: add command `set bitcoin_ibd`
* bbb-config.sh: stop lightningd and electrs on bitcoind resync
* bitcoind: service starts are logged including auth information for debugging


armbian/base/build/customize-armbian-rockpro64.sh
```diff
@@ -242,22 +305,8 @@ Type=simple
 WantedBy=multi-user.target
 EOF
 
-## customize ssh login message
+## disable ssh login messages
```
TheCharlatan: Why remove it?

Stadicus: It's a hassle to adjust it on the readonly file system when switching btw mainnet and testnet and not really needed. We can keep it if you think it's worth it, though.

TheCharlatan: Fine by me to remove it then.

PR #55, opened 2019-06-13
Base Supervisor: initial program outline 

Stadicus:

PR #120, opened 2019-07-25
Add basic functionality to the bbbsupervisor

0xB10C:
The BitBox Base supervisor watches the services running on the BitBox Base and performs actions on certain events. This PR provides a series of commits which cleanup, refactor and build on top of the previous initial commits building the bbbsupervisor (#55, #81). Additionally, this PR is a documentation of existing behavior and how to build upon it.

I noticed that #81 explicitly removed a 3rd party dependency, while this PR (commit b75feeaf6028e7601a7be13a9495ffe24052173a) introduces a dependency on https://github.com/tidwall/gjson to make the handling of unexpected JSON returns from Prometheus easier and future proof. 

## The current state of the bbbsupervisor

This PR implements basic supervisor functionality. Two so-called watchers are implemented. A watcher watches specific resource and triggers events. For some events, actions are defined that are being taken. These two watchers are implemented right now:
- a `logWatcher` watching systemd logs for a specific service (e.g. `bitcoind.service`) via `journalctl`
- a `prometheusWatcher` watching a specific measurement exposed via the Prometheus API

### The `logWatcher`
For each systemd service, a `logWatcher` is started in its own goroutine. It starts to `--follow` the systemd log of that unit via `journalctl`. `stdout` output is written to an `eventWriter` which parses a line (sometimes also multiple lines, known issue) into an event by performing string matching on it. Each `watcherEvent` gets assigned a trigger. The event is passed into a _event channel_ called `events`. `stderr` output is written to an `errWritter` which passes all line(s) read into an _error channel_ called `errs`.

### The `prometheusWatcher`
For each Prometheus value to watch a `prometheusWatcher` is started in its own goroutine. The `prometheusWatcher` queries a specific `measure` or `expression` (not sure why both exist, maybe Stadicus knows more). It passes a watcherEvent into the `events` channel with the `measure` and the measured `value`. The watcher then sleeps and queries again after waking back up. The query `interval` can be set for each `prometheusWatcher`.

### Event handling
Events are indefinitely read from the channels (`errs`, `events`) in the `eventLoop()` function. First errors from the `errs` channel are read (if existent) and a _panic_ is thrown (currently not _recovered_ yet). Then `events` is read and the triggers are handled in the respective handle functions. Then the event handling loop restarts.

### Currently handled event triggers (business logic)

There are currently three triggers handled: `triggerElectrsFullySynced`, `triggerElectrsNoBitcoindConnectivity` and `triggerPrometheusBitcoindIDB`. I propose to document and propose triggers (including action and rationale) in a table.

| trigger | fired when | action performed | rationale |
| ---  | --- | --- | --- |
| `triggerElectrsFullySynced` (logWatcher) | when electrs is fully synced. electrs stdout contains `"finished full compaction"`. | Restart electrs. (this is not tested, just taken from previous versions) | Maybe Stadicus knows? |
| `triggerElectrsNoBitcoindConnectivity` (logWatcher) | when the `.cookie.env` file changes without electrs noticing. electrs stdout `"WARN - reconnecting to bitcoind: no reply from daemon"`` | restart electrs which, via script (7e7a09c5907a68783971a03e52d866b738f75462) recreates the cookie file| electrs sometime gets out of sync with the bitcoind .cookie file |
| `triggerPrometheusBitcoindIDB` (prometheusWatcher) | every time a Prometheus measurement for `"bitcoind_ibd"` has been made | initial trigger: run `bbbconfig.sh set bitcoin_idb true/false`. not changed: nothing, changed: run `bbbconfig.sh set bitcoin_idb true/false` | Setting e.g. a different db_cache and stop lightningd and electrs on initial block download |

For some triggers, a (previous) state is needed. For example `triggerPrometheusBitcoindIDB` needs the previous measurement to detect a change from _idb_ to _no-idb_. For logWatcher triggers, a flood control is implemented. I.e a trigger is only handled again after a definable `minDelay` to prevent multiple handling actions being executed at roughly the same time.

#### Adding a new trigger 

To add a new trigger this procedure can be followed:

1. Add the trigger to the constants and add the name to the `map[trigger]string` called `triggerNames`.
2. When adding a new trigger for a measurement on Prometheus a new `prometheusWatcher` has to be set up in `setupWatchers()`.
3. When adding a new trigger for an existing `logWatcher` a new string matcher has to be added in parseEvent().
4. Add a switch case for the new trigger in `eventLoop()` and add a handling function.
5. Handling functions shouldn't hardcode multiple (more than two) commands. Consider writing a shell script that is being run in the handling function.

### Next steps

Next steps for the supervisor could be (in no particular order):

- If needed create a watcher that checks if a service is not running
- Properly split incoming stdout lines at a `\n`
- As bbbsupervisor.go grows refactor it into multiple files
- Implement proper logging
- Write unit tests for e.g. `isTriggerFlooding()`, `parseEvent()`, ...
- Read `minDelay` for the flood control, query intervals, ... from a config file (maybe a JSON file as in the other Shift projects)
- Implement proper error handling and panic recovery (bbbsupervisor should not crash on an error)
- Handle system signals stopping the execution (e.g. SIGINT, SIGQUIT, SIGTERM)
- Extend `prometheusWatcher.query()` to query for strings, ints ... (currently only `float64`)


Stadicus:
Great work, thanks @0xB10C!

PR #128, opened 2019-08-05
rename base-middleware to bbbmiddleware

Stadicus:
Renaming gives consistency over all our own tools, like bbbsupervisor or bbbfancontrol

### DIFF
--- a/middleware/.gitignore
+++ b/middleware/.gitignore
@@ -1,4 +1,4 @@
-base-middleware
+bbbmiddleware
 
 ### Vim ###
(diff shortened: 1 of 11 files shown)


middleware/.gitignore
```diff
@@ -1,4 +1,4 @@
-base-middleware
+bbbmiddleware
```
TheCharlatan: We can remove this entry completely.

Stadicus: done

PR #132, opened 2019-08-07
add 'exec bitcoin_resync' option to bbb-config.sh

Stadicus:
The command 'bbb-config.sh exec bitcoin_resync' deletes the following directories:
* /mnt/ssd/bitcoin/.bitcoin/blocks
* /mnt/ssd/bitcoin/.bitcoin/chainstate
* /mnt/ssd/electrs/db

It removes the trigger file /data/triggers/bitcoind_fully_synced and restarts bitcoind. The dbcache needs to be set to a higher value manually, or by the supervisor.

### DIFF
--- a/armbian/base/scripts/bbb-config.sh
+++ b/armbian/base/scripts/bbb-config.sh
@@ -300,7 +300,7 @@ case "${COMMAND}" in
 
     exec)
         case "${SETTING}" in
-            BITCOIN_REINDEX)
+            BITCOIN_REINDEX|BITCOIN_RESYNC)
                 # stop systemd services
                 systemctl stop electrs
                 systemctl stop lightningd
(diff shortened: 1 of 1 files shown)


armbian/base/scripts/bbb-config.sh
```diff
@@ -300,7 +300,7 @@ case "${COMMAND}" in
 
     exec)
         case "${SETTING}" in
-            BITCOIN_REINDEX)
+            BITCOIN_REINDEX|BITCOIN_RESYNC)
```
TheCharlatan: Can you add the string to help at the top as well?

Stadicus: Good point, done.

PR #136, opened 2019-08-09
remove local 'theme' to prevent GitHub page build warning

Stadicus:
remove local Jekyll 'theme' to prevent GitHub page build warning

### DIFF
--- a/docs/_config.yml
+++ b/docs/_config.yml
@@ -6,5 +6,7 @@ aux_links:
     "Contribute! Source on GitHub":
       - "https://github.com/digitalbitbox/bitbox-base"
 ga_tracking: UA-133876831-2
-theme: "just-the-docs"
+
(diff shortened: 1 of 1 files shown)


KasparEtter:
I was too late now with my feedback. I don't quite understand yet how all of this works. I'm not sure what https://github.com/digitalbitbox/bitbox-base/blob/632bcd2e2c7931e7a7573fc7131a450ab98d891a/Makefile#L69 does (in particular the `jekyll/jekyll:pages` part). Does this pull in the [GitHub Pages dependency](https://pages.github.com/versions/)? Or how else do you get https://github.com/benbalter/jekyll-remote-theme to run without a Gemfile or `plugins` section in the config? If you have some good answers here, then maybe add them as comments to the config or make-file. Thanks! 🤓

PR #137, opened 2019-08-09
Minor fixes

Stadicus:
Two minor fixes:
* execute systemd-electrs-startpre.sh as root 
* allow incoming TCP port 50002 for electrs in iptables by default  
  Enabling / disabling electrs port should later become part of the configuration management.

### DIFF
--- a/armbian/base/customize-armbian-rockpro64.sh
+++ b/armbian/base/customize-armbian-rockpro64.sh
@@ -571,7 +571,7 @@ PartOf=bitcoind.service
 [Service]
 EnvironmentFile=/etc/electrs/electrs.conf
 EnvironmentFile=/mnt/ssd/bitcoin/.bitcoin/.cookie.env
-ExecStartPre=/opt/shift/scripts/systemd-electrs-startpre.sh
+ExecStartPre=+/opt/shift/scripts/systemd-electrs-startpre.sh
 ExecStart=/usr/bin/electrs \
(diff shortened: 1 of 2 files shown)


armbian/base/customize-armbian-rockpro64.sh
```diff
@@ -571,7 +571,7 @@ PartOf=bitcoind.service
 [Service]
 EnvironmentFile=/etc/electrs/electrs.conf
 EnvironmentFile=/mnt/ssd/bitcoin/.bitcoin/.cookie.env
-ExecStartPre=/opt/shift/scripts/systemd-electrs-startpre.sh
+ExecStartPre=+/opt/shift/scripts/systemd-electrs-startpre.sh
```
TheCharlatan: What does the plus here do?

Stadicus: It runs the `systemd-electrs-startpre.sh` as root, despite the service user set to `electrs`. This is necessary to recreate the `.cookie.env` file and set permissions:

https://github.com/digitalbitbox/bitbox-base/blob/6373a39f7007271f9ccb0911ab341355cf589552/armbian/base/scripts/systemd-electrs-startpre.sh#L19-L21

TheCharlatan: Thanks for the explanation

PR #149, opened 2019-08-20
Fix Armbian Docker build

Stadicus:
Due to recent changes in Armbian Docker build process, this pull-request 
* add the `BUILD_MINIMAL=yes` argument
* installs `curl` for the BitBox Base image

Before moving the final disk image, the number of images available in the target directory is checked and expected to equal 1.


Stadicus:
Tested and built on Ubuntu 18.04 host. Will merge today if nobody cries wolf.

PR #150, opened 2019-08-20
regtest: update regtest setup

Tomasvrba:


middleware/Makefile
```diff
@@ -39,3 +42,10 @@ regtest-down:
 	cd $(REPO_ROOT)/middleware/integration_test ;\
 	docker-compose down ;\
 
+regtest-init:
+	cd $(REPO_ROOT)/middleware/integration_test ;\
+	docker-compose exec bitcoind bitcoin-cli -regtest -rpcport=${RPCPORT} -rpcuser=${RPCUSER} -rpcpassword=${RPCPASSWORD} generate 101
+
+regtest-info:
```
TheCharlatan: The output of this cmand is a bit unintuitive. What is the use you envisioned for it? Would it make more sense for this to return nicely formatted information of all targets?

Tomasvrba: That would probably make more sense, this is just meant to get the info of the bitcoin regtest network, more of a sanity check to see everything is set up correctly. 
Not strictly needed for anything, just a quick way to get the info. Could be renamed to `bitcoin-info` if we keep it?

Otherwise I agree, at some point it would be nice to get info for all the services

PR #155, opened 2019-08-21
Middleware: Add Flashdrive, Backup and Restore RPCs

0xB10C:
**WIP**: First version to gather feedback. Code builds, but I have no BBBase to run it yet.

This PR adds 
- [x]  `Flashdrive` RPC that which handles the args `check`, `mount` and `unmount` 
- [x]  `Backup` RPC that handles the args `sysconfig` and `hsm_secret` 
- [x]  `Restore` RPC that handles the args `sysconfig` and `hsm_secret` 
- [x] rename `usb_flashdrive` to `flashdrive` https://github.com/digitalbitbox/bitbox-base/pull/155#issuecomment-523628270
- [x] Test that the RPCs are callable via the RPC Client in the App

### DIFF
--- a/middleware/src/middleware.go
+++ b/middleware/src/middleware.go
@@ -169,7 +169,7 @@ func (middleware *Middleware) VerificationProgress() rpcmessages.VerificationPro
 func (middleware *Middleware) Flashdrive(args rpcmessages.FlashdriveArgs) (rpcmessages.FlashdriveResponse, error) {
 	switch args.Method {
 	case rpcmessages.Check:
-		log.Println("executing a USB flashdrive check via the cmd script")
+		log.Println("Executing a USB flashdrive check via the cmd script")
 		out, err := middleware.runBBBCmdScript("usb_flashdrive", "check")
 		if err != nil {
 			return rpcmessages.FlashdriveResponse{Success: false, Message: string(out)}, nil
@@ -177,7 +177,7 @@ func (middleware *Middleware) Flashdrive(args rpcmessages.FlashdriveArgs) (rpcme
 		return rpcmessages.FlashdriveResponse{Success: true, Message: string(out)}, nil
 
 	case rpcmessages.Mount:
-		log.Println("executing a USB flashdrive mount via the cmd script")
+		log.Println("Executing a USB flashdrive mount via the cmd script")
 		out, err := middleware.runBBBCmdScript("usb_flashdrive", "mount"+" "+args.Path)
 		if err != nil {
 			return rpcmessages.FlashdriveResponse{Success: false, Message: string(out)}, nil
@@ -185,7 +185,7 @@ func (middleware *Middleware) Flashdrive(args rpcmessages.FlashdriveArgs) (rpcme
 		return rpcmessages.FlashdriveResponse{Success: true, Message: string(out)}, nil
 
 	case rpcmessages.Unmount:
-		log.Println("executing a USB flashdrive unmount via the cmd script")
+		log.Println("Executing a USB flashdrive unmount via the cmd script")
 		out, err := middleware.runBBBCmdScript("usb_flashdrive", "unmount")
 		if err != nil {
 			return rpcmessages.FlashdriveResponse{Success: false, Message: string(out)}, nil
@@ -193,7 +193,31 @@ func (middleware *Middleware) Flashdrive(args rpcmessages.FlashdriveArgs) (rpcme
 		return rpcmessages.FlashdriveResponse{Success: true, Message: string(out)}, nil
 
 	default:
-		return rpcmessages.FlashdriveResponse{Success: false, Message: "FlashdriveMethod not supported. (" + string(args.Method) + ")"}, nil
+		return rpcmessages.FlashdriveResponse{Success: false, Message: "Method " + string(args.Method) + " not supported for Flashdrive."}, nil
+	}
+}
+
+// Backup returns a BackupResponse struct in response to a rpcserver request
+func (middleware *Middleware) Backup(method rpcmessages.BackupArgs) (rpcmessages.BackupResponse, error) {
+	switch method {
+	case rpcmessages.SysConfig:
+		log.Println("Executing a backup of the system config via the cmd script")
+		out, err := middleware.runBBBCmdScript("backup", "sysconfig")
+		if err != nil {
+			return rpcmessages.BackupResponse{Success: false, Message: string(out)}, nil
+		}
+		return rpcmessages.BackupResponse{Success: true, Message: string(out)}, nil
+
+	case rpcmessages.HSMSecret:
+		log.Println("Executing a backup of the c-lightning hsm_secret via the cmd script")
+		out, err := middleware.runBBBCmdScript("backup", "hsm_secret")
+		if err != nil {
+			return rpcmessages.BackupResponse{Success: false, Message: string(out)}, nil
+		}
+		return rpcmessages.BackupResponse{Success: true, Message: string(out)}, nil
+
+	default:
+		return rpcmessages.BackupResponse{Success: false, Message: "Method " + string(method) + " not supported for Backup."}, nil
 	}
 }
 
--- a/middleware/src/middleware.go
+++ b/middleware/src/middleware.go
@@ -136,10 +136,10 @@ func (middleware *Middleware) ResyncBitcoin(option rpcmessages.ResyncBitcoinArgs
 	switch option {
 	case rpcmessages.Resync:
 		log.Println("executing full bitcoin resync in config script")
-		cmd = exec.Command("."+middleware.environment.GetBBBConfigScript(), "exec", "bitcoin_resync")
+		cmd = exec.Command(middleware.environment.GetBBBConfigScript(), "exec", "bitcoin_resync")
 	case rpcmessages.Reindex:
 		log.Println("executing bitcoin reindex in config script")
-		cmd = exec.Command("."+middleware.environment.GetBBBConfigScript(), "exec", "bitcoin_reindex")
+		cmd = exec.Command(middleware.environment.GetBBBConfigScript(), "exec", "bitcoin_reindex")
 	default:
 	}
 	err := cmd.Run()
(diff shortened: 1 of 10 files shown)


Stadicus:
Looks good superficially. For consistency, I could rename the `bbb-cmd.sh` command to `flashdrive` as well (instead of `usb_flashdrive`).

Also, starting to think about a config file probably would make sense as well.

0xB10C:
Resolved linter errors

0xB10C:
> Looks good superficially. For consistency, I could rename the `bbb-cmd.sh` command to `flashdrive` as well (instead of `usb_flashdrive`).

Good point.

> Also, starting to think about a config file probably would make sense as well.
for the middleware I assume?

0xB10C:
I've just noticed that `bbb-cmd.sh usb_flashdrive mount` needs  a `<path>`. That's currently not implemented.

0xB10C:
- [ ] Test that this actually works on the BBBase as expected

Stadicus:
rename ´usb_flashdrive´ to ´flashdrive´ for consistency
https://github.com/digitalbitbox/bitbox-base/pull/157

middleware/src/rpcserver/rpcserver.go
```diff
@@ -51,6 +51,9 @@ func (conn *rpcConn) Close() error {
 type Middleware interface {
 	SystemEnv() rpcmessages.GetEnvResponse
 	ResyncBitcoin(rpcmessages.ResyncBitcoinArgs) (rpcmessages.ResyncBitcoinResponse, error)
+	Flashdrive(rpcmessages.FlashdriveArgs) (rpcmessages.FlashdriveResponse, error)
+	Backup(rpcmessages.BackupArgs) (rpcmessages.BackupResponse, error)
+	Restore(rpcmessages.RestoreArgs) (rpcmessages.RestoreResponse, error)
```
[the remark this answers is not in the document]
0xB10C: Where/what exactly?
https://github.com/digitalbitbox/bitbox-base/pull/155/commits/d4c4993b647bfdd44126b446f76c1df5ae87f3ba#diff-bb48a3164edaae7002e0a4949d26fd9dR225
https://github.com/digitalbitbox/bitbox-base/pull/155/commits/d4c4993b647bfdd44126b446f76c1df5ae87f3ba#diff-11333e3c9d0bc8f26119dc31fdefb911R26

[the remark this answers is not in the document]
TheCharlatan: What calls the interface function here?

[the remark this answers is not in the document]
0xB10C: Oh, good catch! Your right, forgot the actual `Restore` function in `rpcserver.go`. Thanks.

0xB10C:
- Addressed review from @TheCharlatan 
   - Added a unit test to `system_test.go`
   - Added the missing `Restore()` RPC method
- Renamed `usb_flashdrive` to `flashdrive`
- Rebased on master
- Messed up the !fixup

middleware/src/rpcmessages/rpcmessages.go
```diff
@@ -53,3 +88,21 @@ type VerificationProgressResponse struct {
...
+type BackupResponse struct {
+	Success bool
+	Message string
+}
+
+// BackupResponse is the struct that gets sent by the rpc server during a Backup call
+type RestoreResponse struct {
+	Success bool
+	Message string
+}
```
Tomasvrba: Do you think it's necessary to have three different response types for this that are the same?
The whole procedure of plugging in a flashdrive and backing up or restoring is to me part of the same concept so I think could be collapsed into one response type. You already have a message field so additional details could just go there?
 Just a minor nit :)

middleware/src/middleware.go
```diff
@@ -164,3 +164,94 @@ func (middleware *Middleware) SampleInfo() rpcmessages.SampleInfoResponse {
...
 }
+
+// Flashdrive returns a FlashdriveResponse struct in response to a rpcserver request
+func (middleware *Middleware) Flashdrive(args rpcmessages.FlashdriveArgs) (rpcmessages.FlashdriveResponse, error) {
+	switch args.Method {
+	case rpcmessages.Check:
+		log.Println("Executing a USB flashdrive check via the cmd script")
+		out, err := middleware.runBBBCmdScript("flashdrive", "check")
+		if err != nil {
+			return rpcmessages.FlashdriveResponse{Success: false, Message: string(out)}, nil
```
Tomasvrba: Why not return the error as well since you already have it as one of the return types?
Same for other calls below. 
@TheCharlatan I need to catch up with you at some point what kind of error handling you decided on :)

Tomasvrba:
> Nice! I'm happy to see how easy it looks now to add rpc methods. If you want to, you can add the functions to the middleware or rpcserver tests, but it's not that important.

Agreed! Nice work @0xB10C and @TheCharlatan, this looks really awesome to use

middleware/src/rpcmessages/rpcmessages.go
0xB10C: > Do you think it's necessary to have three different response types for this that are the same?

I thought about this as well beforehand, and I don't think it's necessary. Might replace it by a `GenericResponse` which could be used by even more RPCs, than the Flashdrive, Backup and Restore RPCs, that might get added in the future. Good point!

middleware/src/middleware.go
0xB10C: Good point as well. Makes no sense to return `nil` everywhere.

0xB10C:
Addressed @Tomasvrba feedback:
- now using a `GenericResponse` for `Flashdrive()`, `Backup()`, `Restore()` RPCs. The response can be used in the future by other RPCs as well
- the RPC methods now actually return the error and not only send it in the `message` of the `GenericResponse`
- added this to the existing `ResyncBitcoin()` RPC as well in a separate commit f558058. 

(The travis test run fails, because there is now an actual error being returned. Will look into this.)

middleware/src/middleware.go
```diff
@@ -164,3 +166,97 @@ func (middleware *Middleware) SampleInfo() rpcmessages.SampleInfoResponse {
...
+		log.Println("Executing a restore of the c-lightning hsm_secret via the cmd script")
+		out, err := middleware.runBBBCmdScript("restore", "hsm_secret")
+		if err != nil {
+			return rpcmessages.GenericResponse{Success: false, Message: string(out)}, err
+		}
+		return rpcmessages.GenericResponse{Success: true, Message: string(out)}, nil
+
+	default:
+		errorMessage := "Method " + string(method) + " not supported for Restore()."
+		return rpcmessages.GenericResponse{Success: false, Message: errorMessage}, errors.New(errorMessage)
```
TheCharlatan: As weird as it looks, returning the error twice is actually something I'd support. This way we do not need to unpack the error in the rpc call of the wallet app and the frontend twice.

0xB10C:
> (The travis test run fails, because there is now an actual error being returned. Will look into this.)

Refactored the middleware unit tests and added new ones for `Flashdrive()`, `Backup()` and `Restore()`. Additionally I've decided to mock the `bbb-config.sh` and `bbb-cmd.sh` with `/bin/echo`. 

Currently [really struggling](https://travis-ci.org/digitalbitbox/bitbox-base/jobs/576809426#L647) with the `golangci-lint` linter not being happy with a false positive code duplicate. Maybe @Tomasvrba or @TheCharlatan have an idea to circumvent the duplicate detection. I've decided to not spend any more time on this right now.  

```
golangci-lint run
src/middleware.go:205: 205-227 lines are duplicate of `src/middleware.go:230-252` (dupl)
func (middleware *Middleware) Backup(method rpcmessages.BackupArgs) (rpcmessages.GenericResponse, error) {
	switch method {
	case rpcmessages.BackupSysConfig:
		log.Println("Executing a backup of the system config via the cmd script")
		out, err := middleware.runBBBCmdScript("backup", "sysconfig")
		if err != nil {
			return rpcmessages.GenericResponse{Success: false, Message: string(out)}, err
		}
		return rpcmessages.GenericResponse{Success: true, Message: string(out)}, nil
	case rpcmessages.BackupHSMSecret:
		log.Println("Executing a backup of the c-lightning hsm_secret via the cmd script")
		out, err := middleware.runBBBCmdScript("backup", "hsm_secret")
		if err != nil {
			return rpcmessages.GenericResponse{Success: false, Message: string(out)}, err
		}
		return rpcmessages.GenericResponse{Success: true, Message: string(out)}, nil
	default:
		errorMessage := fmt.Sprintf("Method %d not supported for Backup().", method)
		return rpcmessages.GenericResponse{Success: false, Message: errorMessage}, errors.New(errorMessage)
	}
}
src/middleware.go:230: 230-252 lines are duplicate of `src/middleware.go:205-227` (dupl)
func (middleware *Middleware) Restore(method rpcmessages.RestoreArgs) (rpcmessages.GenericResponse, error) {
	switch method {
	case rpcmessages.RestoreSysConfig:
		log.Println("Executing a restore of the system config via the cmd script")
		out, err := middleware.runBBBCmdScript("restore", "sysconfig")
		if err != nil {
			return rpcmessages.GenericResponse{Success: false, Message: string(out)}, err
		}
		return rpcmessages.GenericResponse{Success: true, Message: string(out)}, nil
	case rpcmessages.RestoreHSMSecret:
		log.Println("Executing a restore of the c-lightning hsm_secret via the cmd script")
		out, err := middleware.runBBBCmdScript("restore", "hsm_secret")
		if err != nil {
			return rpcmessages.GenericResponse{Success: false, Message: string(out)}, err
		}
		return rpcmessages.GenericResponse{Success: true, Message: string(out)}, nil
	default:
		errorMessage := fmt.Sprintf("Method %d not supported for Restore().", method)
		return rpcmessages.GenericResponse{Success: false, Message: errorMessage}, errors.New(errorMessage)
	}
}
Makefile:10: recipe for target 'ci' failed
```

A last option would be to disable [dupl](https://github.com/mibk/dupl) (known for false positives) by disabling it via `golangci-lint run --disable=dupl`.

TheCharlatan:
Let's remove it, this is a stupid case. Can we remove it from the golang file?

0xB10C:
> Can we remove it from the golang file?

What do you mean by it? The (false positive) duplicate?

TheCharlatan:
Sorry, I meant this file https://github.com/digitalbitbox/bitbox-base/blob/master/middleware/.golangci.yml

0xB10C:
Increased the `dupl` threshold, ci passes now.

I think this PR is now ready for finial review.

TheCharlatan:
Good job! Did not run any of the code, but looks good to me.

PR #156, opened 2019-08-21
Add Redis for configuration management

Stadicus:
Add Redis to the Armbian build process. It is installed, configured and set up as a systemd unit. Initial configuration is mass-imported from factorysettings.txt

This is the prerequisite to move the configuration management consistently into Redis, which is not yet implemented.

### DIFF
--- a/armbian/base/config/redis/redis.service
+++ b/armbian/base/config/redis/redis.service
@@ -1,0 +1,15 @@
+[Unit]
+Description=Redis In-Memory Data Store
+After=network.target
+# original config in /lib/systemd/system/redis-server.service
+
+[Service]
+Type=forking
+ExecStart=/usr/bin/redis-server /etc/redis/redis.conf
+ExecStop=/bin/kill -s TERM $MAINPID
+User=redis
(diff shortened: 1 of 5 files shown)


armbian/base/config/redis/redis.service
```diff
@@ -0,0 +1,15 @@
+[Unit]
+Description=Redis In-Memory Data Store
+After=network.target
+# original config in /lib/systemd/system/redis-server.service
+
+[Service]
+Type=forking
```
0xB10C: Please test if this works fine when starting and stopping (restarting) Redis on Armbian. (I can't yet.)

I had a problem with `Type=forking`, Redis and systemd before. [Others aswell](https://stackoverflow.com/a/52335904). `Type=notify` could be the right way to go, when `forking` doesn't work.

Stadicus: Current combo (supervised, daemonized + forking) runs fine, but I will try out the option mentioned in this link (supervised + notify) as well. Thanks.

PR #158, opened 2019-08-22
bbbconfgen: add #check option

Stadicus:
The `#check` function allows to drop a line if a key is set to `0`, or not set at all:
```
{{ key #check }}
```

The option can be set anywhere on the line and works together with all other options.


tools/bbbconfgen/test.conf
```diff
@@ -0,0 +1,7 @@
+# network
```
0xB10C: Should this file be added?

tools/bbbconfgen/test/bitcoin-template.conf
```diff
@@ -3,6 +3,7 @@ mainnet={{ bitcoind:mainnet }}
 testnet={{ bitcoind:testnet #default: 0 }}
 rpcconnect={{ bitcoind:rpcconnect }}
 dbcache={{ bitcoind:dbcache #default: 300 }}
-seednode={{ bitcoind:seednode:1 #rmLine }}
-seednode={{ bitcoind:seednode:2 #rmLine }}
-seednode={{ bitcoind:seednode:3 #rmLine }}
+{{ bitcoind:testnet #check }}printtoconsole=1
```
0xB10C: Switch it around i.e.: `printtoconsole=1 {{ bitcoind:testnet #check }}`. Easier to read and consistent with the `seednode` entries below.

tools/bbbconfgen/test.conf
Stadicus: removed

tools/bbbconfgen/test/bitcoin-template.conf
Stadicus: I wanted to explicitly have both options present as an example. But I agree, in the new config files it's like you suggest always, so maybe that's not necessary.

Stadicus: adjusted to your suggestion

PR #159, opened 2019-08-22
Prometheus

Tomasvrba:


Tomasvrba:
This is still work in progress. Had to modify the bitcoind docker image to allow us to run our prometheus python script in there.
My changes to the source are on my fork: https://github.com/Tomasvrba/docker-bitcoin/tree/BBBase and I upload the image to our docker hub org when I make chages: https://hub.docker.com/r/shiftcrypto/basebitcoind/tags
Still some stuff I would like to work out to make this run nicely, but electrs and bitcoind are being already monitored by prometheus

PR #163, opened 2019-08-27
middleware: add hostname RPC with set and get functionality

0xB10C:
This PR adds a `Hostname()` RPC to the middleware. The hostname can either be `set` or `get`.

<s>This PR builds on top of the work done in #155 (only adds a9f5f2f and 17177ea). As soon as that one is merged into master I'll rebase this to master. I'll leave this marked as WIP in the meanwhile. </s>

### DIFF
--- a/middleware/src/middleware.go
+++ b/middleware/src/middleware.go
@@ -6,6 +6,8 @@ import (
 	"fmt"
 	"log"
 	"os/exec"
+	"regexp"
+	"strings"
 	"time"
 
 	"github.com/btcsuite/btcd/rpcclient"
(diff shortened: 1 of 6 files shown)


middleware/src/middleware.go
```diff
@@ -254,6 +254,31 @@ func (middleware *Middleware) Restore(method rpcmessages.RestoreArgs) (rpcmessag
...
+		log.Println("Getting the hostname via the config script")
+		out, err := middleware.runBBBConfigScript("get", "hostname")
+		if err != nil {
+			return rpcmessages.GenericResponse{Success: false, Message: string(out)}, err
+		}
+		return rpcmessages.GenericResponse{Success: true, Message: string(out)}, nil
+
+	case rpcmessages.SetHostname:
+		log.Println("Setting the hostname via the config script")
+		out, err := middleware.runBBBConfigScript("set", "hostname"+" "+args.Name)
```
Tomasvrba: Small nit:
`"hostname"+" "+args.Name` was visually hard for me to parse where all the pluses and quotes belong
Could you do it like in other places please? `"hostname" + " " + args.Name`
I think having spaces in between makes it a lot more readable

0xB10C: Wanted this as well, but the _automatic `gofmt` on saving_ of my IDE changes this back to  `"hostname"+" "+args.Name`... Editing it manually in e.g. `vim` would only help until I edit it next time in the IDE... I'd opt to go with `gofmt` here, even if I don't agree.

0xB10C:
Addressed feedback by @Tomasvrba by passing the second argument and not building strings.
```go
runBBBConfigScript(method string, arg1 string, arg2 string)
```
I think this is cleaner, but not perfectly clean. You still need to pass an empty argument to script calls that only take `arg1`.

0xB10C:
Pushed a commit 057ae817f7e93ee6ffccfaa973cadad75f473fa0 based on recent discussions restricting the hostname to valid chars.

I've based that restriction on a subset of what the [linux man pages define as a valid hostname](http://man7.org/linux/man-pages/man7/hostname.7.html). The Base hostname must be:
- At least two chars long and not longer than 64 chars
- All chars must be lowercase `a-z` or `0-9` or dots or hypens (-), but not as first or last char

This is enforced by the following regex. The regex can be [tested here](https://regex101.com/r/365RAG/4) as well.
```regex
^[a-z0-9]+[a-z0-9-.]{0,62}[a-z0-9]$
```

@Stadicus do we need to change the `bbb-config.sh` script to make this compatible?

middleware/src/middleware.go
```diff
@@ -251,6 +256,37 @@ func (middleware *Middleware) Restore(method rpcmessages.RestoreArgs) (rpcmessag
 	}
 }
 
+// Hostname returns a GenericResponse struct in response to a rpcserver request
+func (middleware *Middleware) Hostname(args rpcmessages.HostnameArgs) (rpcmessages.GenericResponse, error) {
+	switch args.Method {
```
benma: @tblock about `GenericResponse`. I see that it is easier to use when coding, but i would reconsider and start using descriptive structs (and methods!) for everything, as you don't want to deal with a lot of implicit assumptions when reading/maintaining that code.

for example, in this call:

if method==get,genericresponse.Message is the hostname.
if method==set,  what is the Message? need to look at the code, not sure what it is in case of success...


More importantly: we have a great mechanism to add rpc methods, and in this one, you inlining a 2nd layer rpc protocol into one rpc method. This only complicates the code and you need to handle more state (e.g. the default case below). 

Better to just have two mthods: `GetHostname` and `SetHostname` calls, with explicit struct fields.

middleware/src/middleware.go
```diff
@@ -251,6 +256,37 @@ func (middleware *Middleware) Restore(method rpcmessages.RestoreArgs) (rpcmessag
...
+
+		if len(args.Hostname) >= 2 && len(args.Hostname) <= 64 && r.MatchString(args.Hostname) {
+			out, err := middleware.runBBBConfigScript("set", "hostname", args.Hostname)
+			if err != nil {
+				return rpcmessages.GenericResponse{Success: false, Message: string(out)}, err
+			}
+			return rpcmessages.GenericResponse{Success: true, Message: string(out)}, nil
+		}
+		errorMessage := fmt.Sprintf("invalid hostname '%s'", args.Hostname)
+		return rpcmessages.GenericResponse{Success: false, Message: errorMessage}, errors.New(errorMessage)
```
benma: fyi: if you return an error here, the client will get this as an error as well, so you don't need to replicate the error message in the response.

Important: `if err != nil`, then all other return values of a function (or rpc method) should be treated as undefined and not be touched. This makes the caller's life a lot easier.

benma:
Another small thing: GenericResponse.Success is not needed/redundant (it is successful if you don't return an error and not successful if you return an error).

middleware/src/middleware.go
0xB10C: > I see that it is easier to use when coding, but i would reconsider and start using descriptive structs (and methods!) for everything, as you don't want to deal with a lot of implicit assumptions when reading/maintaining that code.

> Better to just have two methods: `GetHostname` and `SetHostname` calls, with explicit input and output structs. 

I agree.

Stadicus:
> This is enforced by the following regex. The regex can be [tested here](https://regex101.com/r/365RAG/4) as well.
> 
> ```regexp
> ^[a-z0-9]+[a-z0-9-.]{0,62}[a-z0-9]$
> ```
> 
> @Stadicus do we need to change the `bbb-config.sh` script to make this compatible?

Will update the script to check this exact regex, see linked issue.

0xB10C:
Split up into `SetHostname` and `GetHostname`. `SetHostname` returns the `ErrorResponse` and `GetHostname` the specific `GetHostnameResponse`. Squashed commits.

0xB10C:
Rebased on master.

0xB10C:

PR #164, opened 2019-08-27
Add dummy RPCs to Authenticate and ChangePassword

0xB10C:
This PR adds two dummy RPCs to authenticate and change the password. This shall only be used for testing and development. 

See https://github.com/digitalbitbox/bitbox-base/pull/164#issuecomment-526148566 for current behavior.

### DIFF
--- a/middleware/src/middleware.go
+++ b/middleware/src/middleware.go
@@ -22,6 +22,10 @@ type Middleware struct {
 	events               chan []byte
 	prometheusClient     *prometheus.PromClient
 	verificationProgress rpcmessages.VerificationProgressResponse
+	// Saves state for the dummy setup process
+	// TODO: should be removed as soon as Authentication is implemented
+	dummyIsBaseSetup   bool
+	dummyAdminPassword string
 }
 
 // NewMiddleware returns a new instance of the middleware
@@ -40,6 +44,8 @@ func NewMiddleware(argumentMap map[string]string) *Middleware {
 			Headers:              0,
 			VerificationProgress: 0.0,
 		},
+		dummyIsBaseSetup:   false,
+		dummyAdminPassword: "",
 	}
 	middleware.prometheusClient = prometheus.NewPromClient(middleware.environment.GetPrometheusURL())
 
@@ -168,6 +174,18 @@ func (middleware *Middleware) VerificationProgress() rpcmessages.VerificationPro
 	return middleware.verificationProgress
 }
 
+// DummyIsBaseSetup returns the current dummyIsBaseSetup bool
+// FIXME: this is a dummy function and should be removed once authentication is implemented
+func (middleware *Middleware) DummyIsBaseSetup() bool {
+	return middleware.dummyIsBaseSetup
+}
+
+// DummyAdminPassword returns the current dummyAdminPassword string
+// FIXME: this is a dummy function and should be removed once authentication is implemented
+func (middleware *Middleware) DummyAdminPassword() string {
+	return middleware.dummyAdminPassword
+}
+
 // Flashdrive returns a GenericResponse struct in response to a rpcserver request
 func (middleware *Middleware) Flashdrive(args rpcmessages.FlashdriveArgs) (rpcmessages.GenericResponse, error) {
 	switch args.Method {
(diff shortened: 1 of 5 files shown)


middleware/src/middleware.go
```diff
@@ -251,6 +251,26 @@ func (middleware *Middleware) Restore(method rpcmessages.RestoreArgs) (rpcmessag
...
+// FIXME: This is a dummy implementation of the Authenticate RPC call
+// in the future this should return an AuthentificationResponse with e.g. an JWT
+func (middleware *Middleware) Authenticate(args rpcmessages.AuthenticateArgs) (rpcmessages.GenericResponse, error) {
+	dummyUsers := map[string]string{
+		"satoshi": "shift1",
+		"sipa":    "shift6",
+		"djb":     "shift9",
+	}
+
+	if expectedPasssword, userIsInMap := dummyUsers[args.Username]; userIsInMap {
```
Tomasvrba: Personal preference to use a custom variable instead of just `ok` which is generally the case in Go?
I don't really mind either way, just wondering :)

Stadicus:
I also think that we should already have the insecure factory passwort built in. I think we also discussed to use `userAuthenticate` and `userChangePassword`.

Our development tests with the app could be:
1. App tries `userAuthenticate` with admin/1234
1. Success, Base is not yet initialized! --> set new password
1. Tester sets to satoshi/shift1 (or any of the above) with `userChangePassword`
1. App is closed and restarted. User authenticates with mallory/evilhack --> Auth fails!
1. App is closed and restarted. User authenticates with satoshi/shift1, sets "remember password"
1. App is closed and restarted. App automatically authenticates with satoshi/shift1

As it is an insecure password, I'd rather have it obiously insecure than to have something cryptic that only seems secure. What do you think about 1234, bitboxbase or CanIHazPassword?

Only change necessary are RPC naming and inclusion of the standard password.

Tomasvrba:
> As it is an insecure password, I'd rather have it obiously insecure than to have something cryptic that only seems secure. What do you think about 1234, bitboxbase or CanIHazPassword?
> 

I definitely vote for "ICanHasPassword?"

0xB10C:
Good feedback, should have looked closer at the actual spec.

middleware/src/middleware.go
0xB10C: Had `ok` at first, but wanted the code to be clear(er), when dealing with passwords. But your right, I'd normally expect a `ok` as well.

0xB10C:
Addressed feedback and added a dummy state as discussed with @Stadicus. Current behavior is as follows:

If the Base is not set up (tracked as the `dummyIsBaseSetup` boolean and reset on middleware restart) you can *only* `UserAuthenticate` with:

| username | password |
|----------|----------|
| admin  | ICanHasPassword?  |

Calling the RPC `UserChangePassword` with the username `admin` let's you choose a new admin password (this is saved in the `dummyAdminPassword` string and reset on middleware restart as well). Changing the admin password sets `dummyIsBaseSetup = true`. It's now possible to login with one of these three:

| username | password |
|----------|----------|
| admin  | (user-set-password)   |
| satoshi     | shift1   |
| dev      | dev  |

The user set password must be at least 8 chars. This can be used to test error behavior when implementing the `userChangePassword` functionality in the front end.

0xB10C:
Rebased onto of #169 and refactored the RPC to return `ErrorResponse` as discussed. Currently not returning an (error) code, since it's a dummy RPC implementation.

0xB10C:
Rebased onto master now.

middleware/src/middleware.go
```diff
@@ -292,10 +310,10 @@ func (middleware *Middleware) UserChangePassword(args rpcmessages.UserChangePass
 				middleware.dummyIsBaseSetup = true // the change of the admin password completes the dummy setup process (for now)
 			}
 		}
+		return rpcmessages.ErrorResponse{Success: true}
 	}
 
-	errorMessage := "authentication unsuccessful"
-	return rpcmessages.GenericResponse{Success: false, Message: errorMessage}, errors.New(errorMessage)
+	return rpcmessages.ErrorResponse{Success: false, Message: "password change unsuccessful (to short)"}
```
Tomasvrba: Typo, "too short"

Tomasvrba: You can also get an ErrorResponse when the provided username is not "admin" right? In that case the message would be incorrect.

0xB10C: No, the `if len(args.NewPassword) >= 8 {...}` block is closed at this point :)

0xB10C:
Squashed and fixed typo.

middleware/src/middleware.go
Tomasvrba: Yep, makes sense. Maybe I looked at a previous commit before or just read it wrong :+1:

PR #167, opened 2019-08-29
Refactor and split up Flashdrive RPC 

0xB10C:
Edit: This PR now addresses two things.

- [x] The Flashdrive RPC calls `bbb-cmd.sh` script with the `check` parameter only internally in `mount`. 
- [x] Split up the Flashdrive RPC into two: `MountFlashdrive` and `UnmountFlashdrive`.

_edit end_

Traffic between the base and the app should be kept minimal. This PR refactors the Flashdrive RPC to only use the `bbb-cmd.sh flashdrive check` output internally when calling the Flashdrive `mount` method. 

The changes make unit testing the RPCs even uglier/more complex. We might need to think about mocking the `bbb-cmd.sh` script in some other way in the future.

### DIFF
--- a/middleware/.golangci.yml
+++ b/middleware/.golangci.yml
@@ -60,7 +60,7 @@ linters-settings:
     suggest-new: true
   dupl:
     # tokens count to trigger issue, 150 by default
-    threshold: 155
+    threshold: 160
   goconst:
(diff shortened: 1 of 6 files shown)


0xB10C:

PR #168, opened 2019-08-30
bbbconfgen: enable #rmLineTrue and #rmLineFalse

Stadicus:
The `#rmLineFalse` function drops a line if the referenced Redis key is empty or false. Based on the (deprecated) `#check` option, is extended to also accept the values 'false', 'no' or 'n' (case insensitive). Both `#rmLineFalse` and `#check` can be used.

The new inverse function `#rmLineTrue` drops the line if the corresponding Redis value is '1', 'true', 'yes' or 'y' (case insensitive).

Tests are adjusted and extended.


Stadicus:
This extension works together with https://github.com/digitalbitbox/bitbox-base/pull/165, but is backwards compatible due to `#check` == `#rmLineFalse` for the moment.

PR #169, opened 2019-08-30
Add ErrorResponse

0xB10C:
This adds an ErrorResponse struct used to idicate if an RPC call was successful or not. It can be embedded into other RPC responses that return values. In any case the ErrorResponse should be checked first, so that, if an error is returned, we ignore everything else in the response.


Tomasvrba:
:+1: 
ack

PR #172, opened 2019-09-03
fix: dummy Authentication RPCs should return an error

0xB10C:
The Go RPC framework requires the RPC call to return an error. This is added for the `UserAuthenticate` and `UserChangePassword` RPCs here.


0xB10C:

PR #175, opened 2019-09-04
minor fixes in backup/restore script and redis systemd unit

Stadicus:
* remove startpre and stop script for Redis systemd unit (was used for read-only filesystem only, now solved)
* fixed backup and restore functions in `bbb-cmd.sh` to preserve file ownership and stop/start Redis correctly


Stadicus:
Rebased and merged

PR #176, opened 2019-09-05
Middleware: Add RPCs needed for the base setup

0xB10C:
This adds the RPCs to the middleware needed to setup the Base in the App. In combination with https://github.com/digitalbitbox/bitbox-wallet-app/pull/506.

### DIFF
--- a/middleware/src/middleware.go
+++ b/middleware/src/middleware.go
@@ -186,36 +186,31 @@ func (middleware *Middleware) DummyAdminPassword() string {
 	return middleware.dummyAdminPassword
 }
 
-// Flashdrive returns a GenericResponse struct in response to a rpcserver request
-func (middleware *Middleware) Flashdrive(args rpcmessages.FlashdriveArgs) (rpcmessages.GenericResponse, error) {
-	switch args.Method {
-	case rpcmessages.Mount:
-		log.Println("Executing a USB flashdrive check via the cmd script")
-		outCheck, errCheck := middleware.runBBBCmdScript("flashdrive", "check", "")
-		if errCheck != nil {
-			return rpcmessages.GenericResponse{Success: false, Message: string(outCheck)}, errCheck
-		}
-		flashDriveName := strings.TrimSuffix(string(outCheck), "\n")
-
-		log.Println("Executing a USB flashdrive mount via the cmd script")
-		outMount, errMount := middleware.runBBBCmdScript("flashdrive", "mount", flashDriveName)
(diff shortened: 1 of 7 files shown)


middleware/src/middleware.go
```diff
@@ -186,87 +184,71 @@ func (middleware *Middleware) DummyAdminPassword() string {
...
-		log.Println("Executing a backup of the c-lightning hsm_secret via the cmd script")
-		out, err := middleware.runBBBCmdScript("backup", "hsm_secret")
-		if err != nil {
-			return rpcmessages.GenericResponse{Success: false, Message: string(out)}, err
-		}
-		return rpcmessages.GenericResponse{Success: true, Message: string(out)}, nil
+// UnmountFlashdrive returns an ErrorResponse struct in a response to a rpcserver request
+func (middleware *Middleware) UnmountFlashdrive() rpcmessages.ErrorResponse {
+	log.Println("Executing a USB flashdrive unmount via the cmd script")
+	out, err := middleware.runBBBCmdScript("flashdrive", "unmount", "")
```
Tomasvrba: `UnmountFlashdrive` doesn't do anything like `middleware.runBBBCmdScript("flashdrive", "check", "")` like `MountFlashdrive` does. What happens if there are multiple drives plugged in, does it unmount all of them?

middleware/src/rpcserver/rpcserver.go
```diff
@@ -83,73 +89,112 @@ func NewRPCServer(middleware Middleware) *RPCServer {
 }
 
 // GetSystemEnv sends the middleware's GetEnvResponse over rpc
-func (server *RPCServer) GetSystemEnv(args int, reply *rpcmessages.GetEnvResponse) error {
+func (server *RPCServer) GetSystemEnv(dummyArg bool, reply *rpcmessages.GetEnvResponse) error {
```
Tomasvrba: TODO for the future: since the rpc server requires args and returning an error, we should write a wrapper so we don't need to pass `dummyArg` and `return nil` everywhere

middleware/src/middleware.go
0xB10C: Good point, but only the drive mounted at `/mnt/backup` is unmounted by the script. 

https://github.com/digitalbitbox/bitbox-base/blob/master/armbian/base/scripts/bbb-cmd.sh#L231-L239

0xB10C:
Added doc strings for `runBBBCmdScript` and `runBBBConfigScript`.

PR #178, opened 2019-09-06
Middleware: update to current bbb-cmd resync|reindex methods

0xB10C:
The `bbb-cmd.sh` methods to reindex and resync where changed in https://github.com/digitalbitbox/bitbox-base/pull/165. This PR restores interoperability between the middleware and the `bbb-cmd.sh` script.


0xB10C:
Good catch!

PR #183, opened 2019-09-11
Bash scripting foundation for Mender over-the-air updates.

Stadicus:
The current firmware version of the build is hardcoded into the textfile /opt/shift/config/version and is not meant to be changed. On boot, it is written into Redis key 'base:version' to access from other programs.

The update process is initiated by the Middleware (not part of this commit) by calling
```
bbb-cmd.sh mender-update install <version>
```
This script checks the version number and calls 'mender -update' with a constructed GitHub releases download link (e.g. https://github.com/digitalbitbox/bitbox-base/releases/download/0.0.2/BitBoxBase-v0.0.2-RockPro64.base)

Redis key 'base:updating' is used to keep track of the update progress. After the update is applied, it is set to 10 and the Middleware restarts the Base.

Valid status codes of 'base:updating'
```
   0: no update in progress
  10: mender update applied
  20: system reconfigured
  30: system tested OK
  40: mender update committed
  90: mender update failed
```
On every boot, the script systemd-update-checks.sh (running as oneshot unit update-checks.service) checks if an update is in progress and executes the following tasks:

* configuration files are recreated and relevant services are restarted --> code 20

* the system is verified, all services must be up and running within 3 minutes --> code 30
  `bbb-systemctl.sh verify`

* on success, the update is committed --> code 40
  `bbb-cmd.sh mender-update commit`

* on failure, a fallback reboot is initiated --> code 50
  `bbb-cmd.sh base restart`

* if update process is finished, reset 'base:updating' to 0


0xB10C:
I have this branch flashed on my base. Running `bbb-cmd.sh mender-update install 0.0.2` fails with following error. Is my overlay root maybe not activated?
 
``` terminal
$ bbb-cmd.sh mender-update install 0.0.2
grep: /etc/overlayroot.local.conf: No such file or directory
grep: /etc/overlayroot.local.conf: No such file or directory
grep: /etc/overlayroot.local.conf: No such file or directory
/usr/local/sbin/bbb-cmd.sh: line 355: mender: command not found
ERR: mender install failed with error code 127
```

Stadicus:
Thanks for testing, it's probably two separate issues:
1. If the image is not built with overlayroot, the file `/etc/overlayroot.local.conf` does not exist. That is detected by the following script (which I would expect to be `-q` quiet): https://github.com/digitalbitbox/bitbox-base/blob/4c4320b511299f56a0827c9e7cc66e7cec11fe20/armbian/base/scripts/include/exec_overlayroot.sh.inc#L5-L8

  It's actually redundant and I can take out the following, but that shouldn't be an issue here: https://github.com/digitalbitbox/bitbox-base/blob/4c4320b511299f56a0827c9e7cc66e7cec11fe20/armbian/base/scripts/bbb-cmd.sh#L50-L54

2. Yes, this action definitely needs to check if Mender is installed. This happens during post-processing with `make mender-artefacts`. Will do that for this pull-request.

Stadicus:
Fixed feedback issues in https://github.com/digitalbitbox/bitbox-base/pull/183/commits/8d274c6ad59bafc0fa57f3e4800ea637f1834657

armbian/base/scripts/systemd-update-checks.sh
```diff
@@ -0,0 +1,119 @@
...
+        if [[ $(redis_get "base:wifi:enabled") -eq 1 ]]; then
+            generateConfig "wlan0.conf.template"
+            echo "INFO: restarted Tor dae and networking daemon"
+        fi
+        echo "INFO: system configuration recreated"
+
+        # restart services (don't exit on failure)
+        if [[ $(redis_get "tor:base:enabled") -eq 1 ]]; then
+            systemctl restart tor.service || true
+            echo "INFO: restarted Tor dae and networking daemon"
```
0xB10C: nit: s/dae/daemon/

armbian/base/scripts/systemd-update-checks.sh
```diff
@@ -0,0 +1,119 @@
...
+        generateConfig "lightningd.conf.template"
+        generateConfig "electrs.conf.template"
+        generateConfig "bbbmiddleware.conf.template"
+        generateConfig "bashrc-custom.template"
+        generateConfig "grafana.ini.template"
+        generateConfig "torrc.template"
+
+        if [[ $(redis_get "base:wifi:enabled") -eq 1 ]]; then
+            generateConfig "wlan0.conf.template"
+            echo "INFO: restarted Tor dae and networking daemon"
```
0xB10C: nit: s/dae/daemon/

Stadicus: fixed

armbian/base/scripts/systemd-update-checks.sh
Stadicus: fixed

0xB10C:
Anything else to test/review? Maybe applying a newer Mender image i.e. master HEAD tagged as 2.0.1?

Stadicus:
I think that's fine for now. The expected version is in the form 0.0.0 and works only for binary releses. Will upload the release and bump the number once we merged most open pull requests.

PR #186, opened 2019-09-11
bbb-config: valdiate hostname using regex

Stadicus:
According to IETF RFC 952 (https://tools.ietf.org/html/rfc952),

    A "name" (Net, Host, Gateway, or Domain name) is a text string up to
    24 characters drawn from the alphabet (A-Z), digits (0-9), minus sign (-),
    and period (.). Note that periods are only allowed when they serve to
    delimit components of "domain style names". No blank or space characters
    are permitted as part of a name. No distinction is made between upper and
    lower case. The first character must be an alpha character. The last
    character must not be a minus sign or period.

Using the following regex to validate hostname before change:

    ^[a-z][a-z0-9-]{0,22}[a-z0-9]$

### DIFF
--- a/armbian/base/scripts/bbb-config.sh
+++ b/armbian/base/scripts/bbb-config.sh
@@ -299,17 +299,17 @@ case "${COMMAND}" in
                 ;;
 
             HOSTNAME)
-                case "${3}" in
-                    [^0-9A-Za-z]*|*[^-0-9A-Z_a-z]*|*[^0-9A-Za-z]|*-_*|*_-*)
-                        echo "Invalid argument: '${3}' is not a valid hostname."
-                        exit 1
-                        ;;
-                    *)
-                        exec_overlayroot all-layers "echo '${3}' > /etc/hostname"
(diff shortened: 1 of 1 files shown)


0xB10C:
To slow.

Stadicus:
From what I understand, a period is only allowed _"when they serve to delimit components of "domain style names"._ 

Do you think that would be a legit hostname? I don't care either way.

0xB10C:
_domain style name_ probably includes hostnames. (_`A "name" (Net, Host, Gateway, or Domain name)`_)

Yet I'm not sure how often people give _domain style names_ to their home devices e.g. `bitcoin.box`...

It makes the regex more complicated and since we are probably deriving the hostname from the name the user gives his BBB this might not need to be changed...

I'd imagine the hostname from name derivation being something along the lines of `toLowercase()` and `replace(" ", "-")`. I don't see how periods would appear there.
```
# name         ->       hostname
Satoshis Base  ->       satoshis-base
Bitcoin Box    ->       bitcoin-box
```

tl;dr: I guess it's fine as is now.

PR #187, opened 2019-09-11
supervisor: add clearnet-ibd finished handling

0xB10C:
When the IBD is finished and the clearnet IBD key is set to 1 (IBD over clearnet) it is set back to 0 (IBD over Tor).

### DIFF
--- a/tools/bbbsupervisor/bbbsupervisor.go
+++ b/tools/bbbsupervisor/bbbsupervisor.go
@@ -72,6 +72,9 @@ var (
 	versionArg   = flag.Bool("version", false, "return program version")
 )
 
+// Redis Connection
+var redisConn redis.Conn
+
 const (
 	helpText = `
 Watches systemd logs (via journalctl) and queries Prometheus to detect potential issues and take action.
@@ -89,7 +92,7 @@ Command-line arguments:
 // e.g. that bitcoin or electrs has fully synced, or a service is not reachable
 type watcherEvent struct {
 	unit    string  // unit represents systemd unit name, e.g. 'bitcoind'
-	trigger trigger // trigger could be e.g. 'triggerElectrsNoBitcoindConnectivity' or 'triggerPrometheusBitcoindIDB'
+	trigger trigger // trigger could be e.g. 'triggerElectrsNoBitcoindConnectivity' or 'triggerPrometheusBitcoindIBD'
 	measure string  // measure is something that is measured by the prometheusWatcher
 	value   float64 // value is the value that has been measured
 }
@@ -127,7 +130,7 @@ type eventWriter struct {
 // the state values are filled over time
 type supervisorState struct {
 	triggerLastExecuted    map[trigger]int64 // implements a state (timestamps) when a trigger was fired (to mitigate trigger flooding)
-	prometheusLastStateIBD float64           // implements a state for the last `bitcoin_ibd` measurement value (to detect switches idb <-> no-idb)
+	prometheusLastStateIBD float64           // implements a state for the last `bitcoin_ibd` measurement value (to detect switches ibd <-> no-ibd)
 }
 
 // trigger is something specific that can happen for a service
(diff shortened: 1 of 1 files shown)


tools/bbbsupervisor/bbbsupervisor.go
```diff
@@ -486,23 +490,46 @@ func handleBitcoindIDB(event watcherEvent, pState *supervisorState) error {
...
-		pState.prometheusLastStateIBD = newValue
+		pState.prometheusLastStateIBD = isActive
 	}
 	return nil
 }
 
+// unsetClearnetIDB unsets (0 - download blocks over Tor) the ibdClearnetRedisKey if set.
+// The key can only be set back to 1 (download blocks over clearnet) via RPC.
+func unsetClearnetIDB() (err error) {
+	const ibdClearnetRedisKey string = "bitcoin:ibd-clearnet"
```
Stadicus: Redis key is `bitcoind:ibd-clearnet`

tools/bbbsupervisor/bbbsupervisor.go
```diff
@@ -486,23 +490,46 @@ func handleBitcoindIDB(event watcherEvent, pState *supervisorState) error {
...
+	isIBDClearnet, err := getRedisInt(ibdClearnetRedisKey)
+	if err != nil {
+		return fmt.Errorf("getting redis key %s failed: %v", ibdClearnetRedisKey, err)
+	}
+	if isIBDClearnet == 1 {
+		log.Printf("IDB finished. Setting %s to %d.\n", ibdClearnetRedisKey, 0)
+		err = setRedisValue(ibdClearnetRedisKey, "0")
```
Stadicus: My bad, it's not enough that the Redis key is set to `0`, also the bitcoind config file must be changed and the service needs to be restarted. The supervisor should call the following command, which also sets the Redis value back to zero.
```
bbb-config.sh disable bitcoin_ibd_clearnet
```
Sorry about the sloppy spec... 😶

0xB10C:
Changed redis key to `bitcoind:ibd-clearnet`, added a `set to 0` if `ìbd==false` on supervisor startup and replaced `setRedisValue` with a `setBBBConfigValue`.

tools/bbbsupervisor/bbbsupervisor.go
```diff
@@ -462,47 +466,75 @@ func handleMiddlewareNoBitcoindConnectivity(event watcherEvent, pState *supervis
...
+// The key can only be set back to 1 (download blocks over clearnet) via RPC.
+func unsetClearnetIDB() (err error) {
+	const ibdClearnetRedisKey string = "bitcoind:ibd-clearnet"
+	isIBDClearnet, err := getRedisInt(ibdClearnetRedisKey)
+	if err != nil {
+		return fmt.Errorf("getting redis key %s failed: %v", ibdClearnetRedisKey, err)
+	}
+	if isIBDClearnet == 1 {
+		log.Printf("IDB finished. Setting %s to %d.\n", ibdClearnetRedisKey, 0)
+		err := setBBBConfigValue("disable", "bitcoin_ibd_clearnet")
```
Stadicus: IIUC, the function `setBBBConfigValue` would call the following script
```
bbb-config.sh set disable bitcoin_ibd_clearnet
```
This would be wrong, as it should be `bbb-config.sh disable bitcoin_ibd_clearnet`. Thinking of it, I could move the command from "enable/disable" section into "set/get". Then this line should be 
```
err := setBBBConfigValue("bitcoin_ibd_clearnet", "false")
```
Should I adjust the bash script?

0xB10C: Yeah you are totally right.. I should check closer.. 

Ok. Let's move it from enable/disable to set/get. I'll fix it here.

0xB10C: If changed, the middleware RPC for that added in 5b6029528e77552dff334555b22f14a131afb7bd needs to be updated as well.

Stadicus: See draft pull-request https://github.com/digitalbitbox/bitbox-base/pull/190 to change bash script

0xB10C:
Now must be merged with #193 and #187.

PR #188, opened 2019-09-11
regtest: update volume mounting path to $HOME to work on MacOS

Tomasvrba:


Tomasvrba:
Closing PR since even with the files mounted on a different path, MacOS can't handle dialing the rpc socket files out of the box from the host machine. 
Dockerizing the middleware is what needs to be done to move forward.

PR #190, opened 2019-09-11
bbb-config: move 'bitcoin_ibd_clearnet' to 'set' section

Stadicus:
new command to configure bitcoind ibd over clearnet is:
```
  bbb-config.sh set bitcoin_ibd_clearnet <true|false>
```


0xB10C:
Making sure the dev-script version stays consistent with the proper scripts is hard and easily forgot on e.g. quick changes like this one. 

Maybe having a flag `--dev` flag like `bbb-config.sh set bitcoin_ibd_clearnet <true|false> --dev` makes things a lot easier. Another way would be to use a distinct environment variable for development.

Stadicus:
Updated dev script for the moment and created issue https://github.com/shiftdevices/bitbox-base-internal/issues/271 to implement development mode for scripts.

PR #191, opened 2019-09-11
middelware: Add EnableRootLogin and SetRootPassword RPCs

0xB10C:
This PR adds the following RPCs

| RPC                | Input                 | Output                 |                  
|--------------------|-----------------------|-----------------------|
| EnableRootLogin    |`enable bool`| ErrorResponse                                       |
| SetRootPassword    | `SetRootPasswordArgs`              | ErrorResponse                               |

The SetRootPassword RPC allows only passwords with a length greater or equal to 8 chars. The rune length is counted, so that Unicode symbols as e.g. `₿` don't count as multiple chars. See https://play.golang.org/p/yw6Adi167tk.


0xB10C:
Rebased onto master.

PR #195, opened 2019-09-16
Middleware: Add redis connectivity

0xB10C:
This PR adds Redis connectivity to the middleware. For now only the two (unused) methods `GetInt(key string)` and `GetString(key string)` are provided.


0xB10C:
Rebased onto master.

PR #200, opened 2019-09-17
Middleware: Add GetBaseVersion RPC

0xB10C:
This PR adds the GetBaseVersion RPC. Additionally a Redis mock is introduced to enable unit tests on RPCs that GET or SET redis values.

### DIFF
--- a/middleware/cmd/middleware/main.go
+++ b/middleware/cmd/middleware/main.go
@@ -46,7 +46,7 @@ func main() {
 		}
 	}
 	defer logBeforeExit()
-	middleware := middleware.NewMiddleware(argumentMap)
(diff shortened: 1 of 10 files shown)


0xB10C:
I'm leaving this as a draft for now until I have tested this on a base.

0xB10C:
> Question: would it make sense the have the `mock` bool as an optional argument with default `false`?

Can't really do that cleanly, because Go does not support optional parameters.

0xB10C:
Successfully tested this with this branch https://github.com/0xB10C/bitbox-wallet-app/tree/2019-09-middleware-getbaseversion-rpc.

PR #201, opened 2019-09-17
Minor fixes / version bump 0.0.3

Stadicus:
Minor changes:
* fix Redis keys for tor config
* add `bbb-config <enable|disable> bitcoin_incoming`
* add redis default keys for `assumevalid`
* fix to execute update scripts after restart even if `base:updating` key did not exist
* add firmware version to Prometheus and Grafana
* bump version to 0.0.3


armbian/base/config/redis/factorysettings.txt
```diff
@@ -14,9 +14,10 @@ SET tor:ssh:enabled 0
 SET tor:electrs:enabled 1
 SET tor:bbbmiddleware:enabled 1
 
-SET bitcoind:version xxx      
+SET bitcoind:version xxx
 SET bitcoind:ibd 1
 SET bitcoind:ibd-clearnet 0
+SET bitcoind:assumevalid 1
```
0xB10C: IIRC Bitcoin Core's `assumevalid` takes a block-hash which is assumed to be valid (i.e. only check the proof of work, no tx signatures before the given block). The current default value is defined [here](https://github.com/bitcoin/bitcoin/blob/00ffe5aca134f424351b05a85da4eef9c3e4553a/src/chainparams.cpp#L97) (Block 563378). I don't see the reason to change the default `assumevalid`.

(is the 1 a placeholder? if yes it should probably be `xxx`)

armbian/base/config/templates/bitcoin.conf.template
```diff
@@ -32,4 +32,5 @@ seednode={{         bitcoind:seednode:2         #rmLine }}                  {{ t
 seednode={{         bitcoind:seednode:3         #rmLine }}                  {{ tor:base:enabled #rmLineFalse }} {{ bitcoind:ibd-clearnet #rmLineTrue }}
 
 # validation
+assumevalid={{        bitcoind:assumevalid         #default: 1 }}
```
0xB10C: see above comment on `assumevalid`.

armbian/base/scripts/bbb-config.sh
```diff
@@ -14,16 +14,17 @@ usage: bbb-config.sh [--version] [--help]
 assumes Redis database running to be used with 'redis-cli'
 
 possible commands:
-  enable    <bitcoin_ibd_clearnet|dashboard_hdmi|dashboard_web|wifi|autosetup_ssd|
+  enable    <bitcoin_incoming|dashboard_hdmi|dashboard_web|wifi|autosetup_ssd|
```
0xB10C: Should `bitcoin_ibd_clearnet` be removed? That seems to be a breaking change.

armbian/base/config/redis/factorysettings.txt
Stadicus: Good find, I was under the impression it's either turned on by default (1) or off. But it's not working like that, I'll remove this change for now.

armbian/base/scripts/bbb-config.sh
Stadicus: This command has been [moved](https://github.com/digitalbitbox/bitbox-base/blob/ac0307e654299102dee9a414f584f2751f4b424c/armbian/base/scripts/bbb-config.sh#L273) from `enable/disable` to the `set` part of the script previously, just updating the usage instructions.

0xB10C: Your right. Did not look properly.

Stadicus:
Reverted `assumevalid` configuration option.

Also: fixed a minor issue that prevented the system reconfiguration when updating from an image that has the Redis key `base:updating` not set.

```
assume update if 'base:updating' is not set
    
    introduces two new error codes:
    * MENDER_UPDATE_SYSCONFIG_FAILED
    * SYSTEMD_NOT_ALL_SERVICES_RUNNING
```

PR #202, opened 2019-09-18
Middleware: Error Codes implementation

0xB10C:
This PR implements Error Codes (based on #198) for the middleware RPCs.

### DIFF
--- a/middleware/src/middleware.go
+++ b/middleware/src/middleware.go
@@ -148,7 +148,7 @@ func (middleware *Middleware) Start() <-chan []byte {
 // ResyncBitcoin returns a ErrorResponse struct in response to a rpcserver request
 func (middleware *Middleware) ResyncBitcoin() rpcmessages.ErrorResponse {
 	log.Println("executing full bitcoin resync via the config script")
-	out, err := middleware.runBBBCmdScript("bitcoind", "resync", "")
+	out, err := middleware.runBBBCmdScript([]string{"bitcoind", "resync"})
 	if err != nil {
 		return rpcmessages.ErrorResponse{Success: false, Message: string(out), Code: err.Error()}
 	}
@@ -158,7 +158,7 @@ func (middleware *Middleware) ResyncBitcoin() rpcmessages.ErrorResponse {
 // ReindexBitcoin returns a ErrorResponse struct in response to a rpcserver request
 func (middleware *Middleware) ReindexBitcoin() rpcmessages.ErrorResponse {
 	log.Println("executing full bitcoin resync via the config script")
-	out, err := middleware.runBBBCmdScript("bitcoind", "reindex", "")
+	out, err := middleware.runBBBCmdScript([]string{"bitcoind", "reindex"})
 	if err != nil {
 		return rpcmessages.ErrorResponse{Success: false, Message: string(out), Code: err.Error()}
 	}
@@ -196,14 +196,14 @@ func (middleware *Middleware) DummyAdminPassword() string {
 // MountFlashdrive returns an ErrorResponse struct in a response to a rpcserver request
 func (middleware *Middleware) MountFlashdrive() rpcmessages.ErrorResponse {
 	log.Println("Executing a USB flashdrive check via the cmd script")
-	outCheck, err := middleware.runBBBCmdScript("flashdrive", "check", "")
+	outCheck, err := middleware.runBBBCmdScript([]string{"flashdrive", "check"})
 	if err != nil {
 		return rpcmessages.ErrorResponse{Success: false, Message: string(outCheck), Code: err.Error()}
 	}
 	flashDriveName := strings.TrimSuffix(string(outCheck), "\n")
 
 	log.Println("Executing a USB flashdrive mount via the cmd script")
-	outMount, err := middleware.runBBBCmdScript("flashdrive", "mount", flashDriveName)
+	outMount, err := middleware.runBBBCmdScript([]string{"flashdrive", "mount", flashDriveName})
 	if err != nil {
 		return rpcmessages.ErrorResponse{Success: false, Message: string(outMount), Code: err.Error()}
 	}
@@ -213,7 +213,7 @@ func (middleware *Middleware) MountFlashdrive() rpcmessages.ErrorResponse {
 // UnmountFlashdrive returns an ErrorResponse struct in a response to a rpcserver request
 func (middleware *Middleware) UnmountFlashdrive() rpcmessages.ErrorResponse {
 	log.Println("Executing a USB flashdrive unmount via the cmd script")
-	out, err := middleware.runBBBCmdScript("flashdrive", "unmount", "")
+	out, err := middleware.runBBBCmdScript([]string{"flashdrive", "unmount"})
 	if err != nil {
 		return rpcmessages.ErrorResponse{Success: false, Message: string(out), Code: err.Error()}
 	}
@@ -223,7 +223,7 @@ func (middleware *Middleware) UnmountFlashdrive() rpcmessages.ErrorResponse {
 // BackupSysconfig returns a ErrorResponse struct in response to a rpcserver request
 func (middleware *Middleware) BackupSysconfig() rpcmessages.ErrorResponse {
 	log.Println("Executing a backup of the system config via the cmd script")
-	out, err := middleware.runBBBCmdScript("backup", "sysconfig", "")
+	out, err := middleware.runBBBCmdScript([]string{"backup", "sysconfig"})
 	if err != nil {
 		return rpcmessages.ErrorResponse{Success: false, Message: string(out), Code: err.Error()}
 	}
@@ -233,7 +233,7 @@ func (middleware *Middleware) BackupSysconfig() rpcmessages.ErrorResponse {
 // BackupHSMSecret returns a ErrorResponse struct in response to a rpcserver request
 func (middleware *Middleware) BackupHSMSecret() rpcmessages.ErrorResponse {
 	log.Println("Executing a backup of the c-lightning hsm_secret via the cmd script")
-	out, err := middleware.runBBBCmdScript("backup", "hsm_secret", "")
+	out, err := middleware.runBBBCmdScript([]string{"backup", "hsm_secret"})
 	if err != nil {
 		return rpcmessages.ErrorResponse{Success: false, Message: string(out), Code: err.Error()}
 	}
@@ -243,7 +243,7 @@ func (middleware *Middleware) BackupHSMSecret() rpcmessages.ErrorResponse {
 // RestoreSysconfig returns a ErrorResponse struct in response to a rpcserver request
 func (middleware *Middleware) RestoreSysconfig() rpcmessages.ErrorResponse {
 	log.Println("Executing a restore of the system config via the cmd script")
-	out, err := middleware.runBBBCmdScript("restore", "sysconfig", "")
+	out, err := middleware.runBBBCmdScript([]string{"restore", "sysconfig"})
 	if err != nil {
 		return rpcmessages.ErrorResponse{Success: false, Message: string(out), Code: err.Error()}
 	}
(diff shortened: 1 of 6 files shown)


0xB10C:
> Does it make sense to provide all expected error codes when calling an external command? This seems to add a lot of lines and prevent encapsulating the whole error handling in a more generic function once.

Not sure how a generic error handling function would look like. But I'll definitely think that trough once more from different angles. 

> Maybe include the following two error codes from #201 (comment) as well? These are not returned to the Middleware as the script is run by systemd after update reboot, but can be detected within the journald log in a future iteration.


I'd guess that we need the supervisor watching the journald log for `MENDER_UPDATE_SYSCONFIG_FAILED` then? That should write a flag somewhere for the middleware to query system status. 

At the moment middleware and supervisor don't share any code, yet the App and middleware share code. So I'd rather just add the ErrorCodes that are needed for the middleware and app to avoid confusion.

 Something like  `SYSTEMD_NOT_ALL_SERVICES_RUNNING` seems to be needed as a possible response from a GetStatus RPC. But I think there should be defined `StatusCode`s for that (and not use `ErrorCode`s.

0xB10C:
Once #200 is merged this should be rebased and an ErrorCode for a Redis Error should be introduced. Currently there is only only a placeholder as error code.

0xB10C:
> Once #200 is merged this should be rebased and an ErrorCode for a Redis Error should be introduced. Currently there is only only a placeholder as error code.

last push rebased ontop of work done in #200 and introduced an `ErrorCode` for Redis actions.

0xB10C:
> 1. Does it make sense to provide all expected error codes when calling an external command? This seems to add a lot of lines and prevent encapsulating the whole error handling in a more generic function once.

Moved the _default/often occurring_ ErrorCodes into `handleBBBScriptError()`.

middleware/src/middleware.go
```diff
@@ -522,37 +668,100 @@ func (middleware *Middleware) SetRootPassword(args rpcmessages.SetRootPasswordAr
...
-	cmdAsString := script + " " + strings.Join(args, " ")
-	out, err = exec.Command(script, args...).Output()
+func (middleware *Middleware) runBBBConfigScript(args []string) (outputLines []string, err error) {
+	outputLines, err = runCommand(middleware.environment.GetBBBConfigScript(), args)
+	return
+}
+
+// runCommand runs the passed command and returns the combined stdout and stderr output.
+// If the command could not be run, err is not nil.
+func runCommand(command string, args []string) (combinedLines []string, err error) {
```
Tomasvrba: What about typing the `command string` parameter instead with something like `type BBBScript` which could be either `BBBCmdScript` or `BBBConfigScript` so `runCommand` only accepts predefined scripts instead of any string?

middleware/src/middleware.go
```diff
@@ -522,37 +668,100 @@ func (middleware *Middleware) SetRootPassword(args rpcmessages.SetRootPasswordAr
...
+
+	combined := strings.TrimSuffix(string(rawstdoutStderr), "\n")
+	combinedLines = strings.Split(combined, "\n")
+	return combinedLines, err
+}
+
+func handleBBBScriptError(outputLines []string, err error, possibleErrors []rpcmessages.ErrorCode) rpcmessages.ErrorCode {
+	// There are two possible types of errors handled here:
+	// 1.   The script could not be found.
+	// 2.   Script exited with `exis status 1`:
```
Tomasvrba: typo `exit`

middleware/src/middleware.go
```diff
@@ -522,37 +668,100 @@ func (middleware *Middleware) SetRootPassword(args rpcmessages.SetRootPasswordAr
...
 		}
+
 		return rpcmessages.ErrorResponse{Success: true}
 	}
-	return rpcmessages.ErrorResponse{Success: false, Message: "invalid password"}
+
+	return rpcmessages.ErrorResponse{
+		Success: false,
+		Message: "The password has to be at least 8 chars. An unicode char is counted as one.",
+		Code:    rpcmessages.ErrorSetRootPasswordToShort,
```
Tomasvrba: typo `ErrorSetRootPasswordTooShort`

middleware/src/middleware.go
```diff
@@ -522,37 +668,100 @@ func (middleware *Middleware) SetRootPassword(args rpcmessages.SetRootPasswordAr
...
+		log.Printf("Error executing command '%s %s': '%s'", command, strings.Join(args, " "), err.Error())
 	}
-	return
+
+	combined := strings.TrimSuffix(string(rawstdoutStderr), "\n")
+	combinedLines = strings.Split(combined, "\n")
+	return combinedLines, err
+}
+
+func handleBBBScriptError(outputLines []string, err error, possibleErrors []rpcmessages.ErrorCode) rpcmessages.ErrorCode {
```
Tomasvrba: I'm not sure about the `possibleErrors []rpcmessages.ErrorCode` parameter. 
This means that anyone implementing new RPC calls will have to know about all the possible errors and pass the appropriate ones to `handleBBBScriptError`m this seems like a lot of overhead for anyone not familiar with the errors, and a lot of manual checking to make sure you included all the necessary errors, which might also change if we add more errors/codes in the future

Tomasvrba:
Nice wort @0xB10C, very thorough :)
Apart from the typos I have one main request about the `possibleErrors []rpcmessages.ErrorCode` parameter. Maybe I'm wrong, I only scanned through everything pretty quickly, but please take a look at my comment and see if that error handling could be improved (haven't looked at it enough atm to suggest a good solution), or if I am wrong about the issue I described in the comment. Thanks! :)

middleware/src/middleware.go
0xB10C: I see your point, but I think somebody adding an RPC that executes a script should need to know what the script does and which errors it can throw (either by checking or having a spec from the script author). It's currently designed that `possibleErrors` only needs to take the exact error that the  _method_ invoked in the script can throw. All other possible errors like e.g. `ErrorScriptNotSuperuser` are checked by `handleBBBScriptError()`. The `possibleErrors` approach has the benefit that if the script or the code don't match an `ErrorUnhandled` ("UNKNOWN") is returned. That should be something never occurring besides when there is an implementation error in either the the RPC or the scripts. 

(I'll add some of this as a comment though)

middleware/src/middleware.go
0xB10C: To replace `runBBBCmdScript()` and `runBBBConfigScript()` ?

0xB10C:
Pushed: typo fixes, added more code documentation and replaced some `if else` stuff with a `switch-case`

0xB10C:
found and fixed some more 'to short' -> 'too short' typos...

middleware/src/middleware.go
Tomasvrba: Sorry, should have been clearer, what I meant was something like this:
`func runCommand(command BBBscript, args []string) (combinedLines []string, err error) {...}`
but then I suppose we would need `GetBBBConfigScript()` to return `type BBBScript string` first anyway. My thought was to just make the parameters `runCommand()` accepts more specific. But let's just say that's out of scope for this PR. Something to keep in mind for future refactoring

middleware/src/middleware.go
Tomasvrba: I would still really disagree on this. Passing possible error codes to a function just adds an extra layer of indirection. If there is a function such as `handleBBBScriptError()` it should handle all errors, at least that's what I would expect as the caller. If it accepts possible errors as arguments, you might as well handle those possible errors before even calling `handleBBBScriptError()` since you already know what those errors are. Still not ideal though. I actually think this relates nicely to Kaspar's design by contract. After receiving an error from the base and calling `handleBBBScriptError()` I would expect it to return the appropriate error code. Having the additional responsibility as the caller to pre-define some of the function's possible return values just feels really really strange.

middleware/src/middleware.go
```diff
@@ -521,38 +667,105 @@ func (middleware *Middleware) SetRootPassword(args rpcmessages.SetRootPasswordAr
...
```
Tomasvrba: Oh, could you also rename `handleBBBScriptError()` to something more like `handleBBBScriptErrorCode()` since it's just for returning `ErrorCode`

middleware/src/middleware.go
0xB10C: What do you think about having the error codes in a `map[string]ErrorCode` and doing lookups in `handleBBBScriptErrorCode()` ? _(I not sure if a  `map[string]ErrorCode` would be the best datatype for that. We probably don't want to add every new `ErrorCode` to it manually ... Will need to think about that a bit more.)_ The downside is that we wouldn't know when the script returns a not valid error code for that script call (e.g. `flashdrive check` returns `SET_BITCOINETWORK_INVALID_VALUE` and we happily pass that back as a the RPC ErrorCode).

Tomasvrba: Can you give an example for when/why a script would return an invalid error code for that script call? Then it's the responsibility of the script to return the appropriate code and the middleware shouldn't be sanitizing what the scripts return. Why would for example `flashdrive check` return `SET_BITCOINETWORK_INVALID_VALUE`? That seems to defeat the purpose of returning appropriate error codes from the scripts if you can't rely on them and just do separate lookups/sanity checks in the middleware again. 

I suppose my ideal scenario is that since the scripts should return informative, unique and appropriate error codes that the middleware can just pass them on to the app which can then handle the lookup since it needs to do translation for each possible error anyway, otherwise give an unknown error if it doesn't know how to handle it. But maybe I'm missing something why this is not possible/good solution.

benma: > I would expect it to return the appropriate error code. Having the additional responsibility as the caller to pre-define some of the function's possible return values just feels really really strange.

But it's not a "caller", it's all in the same package/file, so it's an implementation detail where you define the valid response codes per api call.

It does make sense to specify clearly which errors an api endpoint can return, and not accidentally "leak" more error codes as you introduce them in other parts of the code.

Imho "Unhandled" is just misleading naming, better would be "Other" or "Unexpected".

Tomasvrba: > But it's not a "caller", it's all in the same package/file, so it's an implementation detail where you define the valid response codes per api call.
> 
> It does make sense to specify clearly which errors an api endpoint can return, and not accidentally "leak" more error codes as you introduce them in other parts of the code.

Yes, agreed that specifying the errors makes sense, I just wouldn't do it in explicitly in every `handleBBBScriptError()` call. An external lookup table, something like the `map[string]ErrorCode` map Timo is proposing would make more sense I think. That way when you call `handleBBBScriptError()` the implementation can already do all the error code handling.

middleware/src/middleware.go
```diff
@@ -521,38 +667,105 @@ func (middleware *Middleware) SetRootPassword(args rpcmessages.SetRootPasswordAr
...
+	// There are two possible types of errors handled here:
+	// 1.   The script could not be found.
+	// 2.   Script exited with `exit status 1`:
+	// 	2.1. Script was not run as superuser. ErrorCode ErrorScriptNotSuperuser is expected as the last outputLine.
+	// 	2.2. CMD Script was not run with correct parameters. ErrorCode ErrorCmdScriptInvalidArg is expected as the last outputLine.
+	// 	2.3. Config Script was not run with correct parameters. ErrorCode ErrorConfigScriptInvalidArg is expected as the last outputLine.
+	// 	2.4. One of the `possibleErrors` is expected as the last outputLine.
+	// All other errors are unknow and not handled. ErrorUnhandled is returned as a last resort.
+
+	if strings.Contains(err.Error(), "no such file or directory") {
```
0xB10C: As @benma pointed out: Don't do string matching here. 

I'll match an `os.ErrNotExist` error.

0xB10C:
This PR should implement the newly added `SCRIPT_INCLUDES_NOT_FOUND` as well. https://github.com/digitalbitbox/bitbox-base/pull/212#issuecomment-535954964

middleware/src/middleware.go
0xB10C: Offline decision: leave as is and iterate on in the future if needed.

PR #203, opened 2019-09-19
scripts: cleanup / add error codes

Stadicus:
all scripts have been cleaned up with shellcheck, unnecessary parts are removed and all aborts are handled with an error code.

The following new error codes are introduced:

generic
* SCRIPT_INVALID_ARG

ssd autosetup
* SSDSETUP_NO_TARGET
* SSDSETUP_TOO_MANY_TARGETS
* SSDSETUP_ABORTED_MANUALLY

lightningd / electrs
* BITCOIND_DEPENDENCY_NOT_ACTIVE
* BITCOIND_IN_IBD

bbb-systemctl.sh
* SYSTEMD_NOT_ALL_SERVICES_RUNNING

systemd-bitcoin-startpre.sh
* BITCOIND_DIRECTORY_NOT_ACCESSIBLE

redis.sh.inc
* REDIS_REQUIRED_BUT_NOT_RUNNING

systemd-startup-checks.sh
* SSD_NOT_MOUNTED

systemd-update-checks.sh
* MENDER_UPDATE_SYSCONFIG_FAILED

### DIFF
--- a/armbian/base/rootfs/etc/systemd/system/bitcoind.service
+++ b/armbian/base/rootfs/etc/systemd/system/bitcoind.service
@@ -14,7 +14,6 @@ ExecStartPre=+/opt/shift/scripts/systemd-bitcoind-startpre.sh
 ExecStart=/usr/bin/bitcoind \
     -conf=/etc/bitcoin/bitcoin.conf \
     -pid=/run/bitcoind/bitcoind.pid
-ExecStartPost=/opt/shift/scripts/systemd-bitcoind-startpost.sh
 
 # Process management
 ####################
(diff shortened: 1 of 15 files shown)


0xB10C:
I like the effort. Make script failures and logs cleaner, but we should probably not confuse App<->Middleware RPC ErrorCode`s with system error codes that probably won't be transmitted to the App in a response to a RPC call faliure.

Stadicus:
Do you think we should treat these errors differently, or use a prefix or something? Or not have these error codes at all?

0xB10C:
More just fyi that I see no point in implementing them in e.g. the middleware `errorcodes.go`. 

If we wan't to parse logs, then these probably will work best/easiest. Not sure when and where that parsing should happen though.

Stadicus:
Ok, thanks. Waiting for your official approval before merging.

PR #208, opened 2019-09-22
iptables: dynamic config from template

Stadicus:
To avoid permissive iptables rules, the various configurations (esp. `tor:base:enabled` and `bitcoind:ibd-clearnet`) include or exclude certain rules when recreating the file `iptables.rules` from the template `iptables.rules.template`.

The build script and especially `bbb-config.sh` is extended to recreate and activate these rules by triggering the `iptables-restore.service` via systemd. This service no longer calls `/sbin/iptables-restore` directly but the script `systemd-iptables-restore.sh` that first flushes existing iptables rules.

Minor fix: also cleanup loop devices before mender-convert if trigger file is present (not just after).

### DIFF
--- a/armbian/base/config/redis/factorysettings.txt
+++ b/armbian/base/config/redis/factorysettings.txt
@@ -52,6 +52,7 @@ SET lightningd:log-level debug
 SET lightningd:plugin:1 /opt/shift/scripts/prometheus-lightningd.py
 
 SET electrs:version xxx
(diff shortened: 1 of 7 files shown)


Stadicus:
Agreed. I tried this manually switching Tor / Clearnet IBD on/off repeatedly, but a real stress-test coming from the App / Middleware would be helpful once ready.

PR #211, opened 2019-09-24
Middleware: Include Flashdrive methods in Backup/Restore Sysconfig

0xB10C:
The middleware calls the bbb-cmd script with `check` `mount` and `unmount` for flashdrive operations. These fuctions were avaliable as RPCs, but are now included in the Backup and Restore sysconfig RPCs. This is done to minimize the amount of RPC calls and traffic, which might need to be send over Tor.

Removed are following RPCs:
- `MountFlashdrive`
- `UnmountFlashdrive`


0xB10C:
Rebased on top of master

PR #212, opened 2019-09-24
MockMode for scripts

Stadicus:
https://github.com/shiftdevices/bitbox-base-internal/issues/271

Because:

* During development and integration tests, the scripts 'bbb-config.sh'
  and 'bbb-cmd.sh' fail as not the whole environment is available.
* Previously, a different development version of these scripts was kept,
  but that is hard to maintain over time.
* By extending the real scripts with a MockMode (set by the environment
  variable MOCKMODE), they can be called for testing purposes while
  still doing all input validation and sanity checks.

This commit:

* extends the two scripts with the functionality to call them in MockMode
* removes the development copies in /middleware/integration_test


Stadicus:
The error SCRIPT_NOT_RUN_AS_SUPERUSER is thrown on purpose to have the exact same conditions as when running on the Base.

If that is not what is expected for integration testing, I can remove that constraint.

0xB10C:
Offline decision: do not echo `SCRIPT_NOT_RUN_AS_SUPERUSER` in MockMode

Stadicus:
Disabled the need for superuser execution in MockMode.
```
$ ./bbb-config.sh enable tor
./bbb-config.sh: needs to be run as superuser.
SCRIPT_NOT_RUN_AS_SUPERUSER

$ MOCKMODE=1 ./bbb-config.sh enable tor
MOCK MODE enabled
OK: enable -- TOR

$ MOCKMODE=1 ./bbb-cmd.sh setup datadir
MOCK MODE enabled
OK: SETUP -- DATADIR -- 
```

Stadicus:
Added new error code: `SCRIPT_INCLUDES_NOT_FOUND`

Long text:
> ERR: includes directory /opt/shift/scripts/include/ not found, must run on BitBox Base system. Run in MockMode for testing.

0xB10C:
> Added new error code: `SCRIPT_INCLUDES_NOT_FOUND`

Can/should this be added to the middleware Error handling as well?

Stadicus:
If it's not too big of a hassle it would be good, as it indicates that there is something seriously wrong with the environment and the script will break in unexpected ways if the includes are not available.

Stadicus:
Rebased on master, merging now

PR #214, opened 2019-09-26
middleware: handle binary setting calls with rpcmessage, not bool

Tomasvrba:
This commit uses the new ToggleSetting type to handle enabling or
disabling Base settings that have binary options.
This is for two reasons:
1. All parameters used in rpc calls should come from rpcmessages.go as discussed with @TheCharlatan 
2. Simplify middleware methods so there is no need for global
"enable/disable" variables and checking within each method call

### DIFF
--- a/middleware/src/middleware.go
+++ b/middleware/src/middleware.go
@@ -31,14 +31,6 @@ type Middleware struct {
 	dummyAdminPassword string
 }
 
-const (
-	// Since enable and disable are two often-used parameters passed to bbb-config
-	// and golangci-lint fails because of "string `disable` has 3 occurrences, make it a constant (goconst)"
-	// they are constants.
-	enableAction  string = "enable"
-	disableAction string = "disable"
-)
-
 // NewMiddleware returns a new instance of the middleware.
 // For testing a mock boolean can be passed, which mocks e.g. redis.
 func NewMiddleware(argumentMap map[string]string, mock bool) *Middleware {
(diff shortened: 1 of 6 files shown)


Stadicus:
Changed the two configuration options `bitcoin_ibd` and `bitcoin_ibd_clearnet` to be "enabled" or "disabled" like the other boolean settings, instead of "set true" and "set false".

Script is backwards compatible with the deprecated settings.

https://github.com/digitalbitbox/bitbox-base/pull/216

middleware/src/rpcmessages/rpcmessages.go
```diff
@@ -40,6 +40,15 @@ type SetRootPasswordArgs struct {
 	RootPassword string
 }
 
+// ToggleSetting is a generic message for settings that can be enabled or disabled
+type ToggleSetting string
+
+// Consts for toggling a setting which is done by passing "enable" or "disable" to the corresponding shell script
+const (
+	ToggleSettingENABLE  ToggleSetting = "enable"
+	ToggleSettingDISABLE ToggleSetting = "disable"
```
0xB10C: Specific reason to have `ENABLE` and `DISABLE` in uppercase? I'd rather stick to the camelcase used usually in Go. e.g. `ToggleSettingEnable` and `ToggleSettingDisable`

middleware/src/middleware.go
```diff
@@ -354,91 +346,63 @@ func (middleware *Middleware) GetHostname() rpcmessages.GetHostnameResponse {
...
-	var action string
-	if enable {
-		log.Println("Enabling Tor via the config script")
-		action = enableAction
-	} else {
-		log.Println("Disabling Tor via the config script")
-		action = disableAction
-	}
+func (middleware *Middleware) EnableTor(toggleAction rpcmessages.ToggleSetting) rpcmessages.ErrorResponse {
+	log.Println("Executing " + toggleAction + " Tor via the config script")
```
0xB10C: That should probably be a `log.Printf` for consistency? *(multiple concurrencys)*
```suggestion
	log.Printf("Executing a '%s Tor' via the config script.\n",  toggleAction)
```

Tomasvrba: Sure. I suppose you mean that for all the methods in this PR then?

middleware/src/rpcmessages/rpcmessages.go
Tomasvrba: Yep, I'll change it. 
Just habit from BB02 where the protobuf messages are generated like that.

Tomasvrba:
> Changed the two configuration options `bitcoin_ibd` and `bitcoin_ibd_clearnet` to be "enabled" or "disabled" like the other boolean settings, instead of "set true" and "set false".
> 
> Script is backwards compatible with the deprecated settings.
> 
> #216

Awesome, thanks Roland. Will update this PR to reflect those changes

Tomasvrba:
> Concept ACK.
> 
> fyi: I did originally design it be booleans to minimize bytes needed to be transferred. I guess that's not to big of a deal here. Minimizing 'messages sent' is probably a priority rather then '+- 10 bytes in a message'.

Yep, totally agree and it absolutely makes sense the way you did it. As you said, this is more about consistency in the communication protocol and being explicit about what types of messages each method requires than about a few extra bytes.

middleware/src/middleware.go
0xB10C: Yes

PR #215, opened 2019-09-26
prometheus: add local ip address

Stadicus:
https://github.com/shiftdevices/bitbox-base-internal/issues/284

Because:

* local IP address can change and needs to be updated frequently. It
  makes sense to add it to the prometheus-base.py scraper, so that
  this value becomes available in Prometheus, ready to be queried by
  other programs.

This commit:

* extens the 'prometheus-base.py' scraper to add the ip_address to
  the the base_system_info object
* always returns the IP address that is used to actively query
  external addresses.
* the script connects to a dummy IP address (192.0.2.0) that is
  reserved by IETF RFC 5737 for documentation purposes and should
  be blocked by ISPs
  https://datatracker.ietf.org/doc/rfc5737/

Output of scraper:
```
base_system_info{base_hostname="bitbox-base",base_ipaddress="192.168.1.161",base_version="0.0.3",build_commit="4c057b3",build_date="2019-09-21",build_time="22:37"} 1.0
```

### DIFF
--- a/armbian/base/scripts/prometheus-base.py
+++ b/armbian/base/scripts/prometheus-base.py
@@ -1,7 +1,7 @@
 #!/usr/bin/env python3
 # -*- coding: utf-8 -*-
 #
-# This script is called by the prometheus-base.service 
+# This script is called by the prometheus-base.service
 # to provide system metrics to Prometheus.
 #
 
@@ -9,6 +9,7 @@ import json
 import time
 import subprocess
 import sys
+import socket
 import redis
 from prometheus_client import start_http_server, Gauge, Counter, Info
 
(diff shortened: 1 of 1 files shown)


0xB10C:
> Output of scraper:
```
base_system_info{base_hostname="...", base_ipaddress="192.168.1.161",..."} 1.0
```

I haven't looked into the Prometheus stuff deeper yet, but would it be possible to have a query like `base_local_ip` that returns just the IP and not `1.0` as a value? (I don't know if the Prometheus API can only return floats, if so I'll find a way to read it from the _returned query_)

Stadicus:
Prometheus is made to track (numeric) metrics. Other data is stored in the INFO type, which is always a list of key/value items.

With `bitcoin_tor_address`, it's just a list of one item. Here, `base_ipaddress` is an additional item in the existing list `base_system_info`. 

I can make a separate list, if needed, but that does not change how the value is wrapped into a list. It's then just the single item.

PR #217, opened 2019-09-26
Minor fixes and improvements

Stadicus:
Collection of little fixes and improvements:
* mDNS: broadcast version as metadata with avahi
* mender: check if Mender is enabled before initializing `/etc/mender` on startup
* mender: add update from flashdrive option
* bitcoind: remove .onion seednodes, as these are updated and built into the bitcoind binaries
* prometheus-bitcoind: fix typo for tor address (`bitcoin_tor_address` is correct)
* scripts: remove 'shutdown' and 'restart' commands from `bbb-cmd.sh`
* scripts: no error output when image built without overlayroot
* cleanup: remove trailing whitespace and fix EOF linebreaks


TheCharlatan:
As long as you are fixing dangling whitespaces, can you squash the ones in build.conf as well?

PR #218, opened 2019-09-30
middleware: add first version of GetBaseInfo RPC

0xB10C:
The GetBaseInfo RPC provides information about the Base, which is displayed in the app's 'Base management' section. The RPCs `GetHostname` and `GetBaseVersion` are included in `GetBaseInfo`.

This PR includes placeholder values for:
- the current Base system status
- the Middleware's hidden service tor port

This PR does not include a unit test for `GetBaseInfo`, because a Prometheus mock is needed first.

I'll keep track of both, the placeholders (https://github.com/shiftdevices/bitbox-base-internal/issues/292) and unit tests (https://github.com/shiftdevices/bitbox-base-internal/issues/274) in a separate issues.

### DIFF
--- a/middleware/cmd/middleware/main.go
+++ b/middleware/cmd/middleware/main.go
@@ -16,6 +16,7 @@ import (
 const version string = "0.0.1"
 
 func main() {
+	middlewarePort := flag.String("middlewareport", "8845", "Port the middleware should listen on (default 8845)")
 	bitcoinRPCUser := flag.String("rpcuser", "rpcuser", "Bitcoin rpc user name")
 	bitcoinRPCPassword := flag.String("rpcpassword", "rpcpassword", "Bitcoin rpc password")
 	bitcoinRPCPort := flag.String("rpcport", "18332", "Bitcoin rpc port, localhost is assumed as an address")
@@ -30,6 +31,7 @@ func main() {
 	flag.Parse()
 
 	argumentMap := make(map[string]string)
+	argumentMap["middlewarePort"] = *middlewarePort
 	argumentMap["bitcoinRPCUser"] = *bitcoinRPCUser
 	argumentMap["bitcoinRPCPassword"] = *bitcoinRPCPassword
 	argumentMap["bitcoinRPCPort"] = *bitcoinRPCPort
@@ -55,9 +57,9 @@ func main() {
 	log.Println("--------------- Started middleware --------------")
 
 	handlers := handlers.NewHandlers(middleware, *dataDir)
-	log.Println("Binding middleware api to port 8845")
+	log.Printf("Binding middleware api to port %s\n", *middlewarePort)
 
-	if err := http.ListenAndServe(":8845", handlers.Router); err != nil {
+	if err := http.ListenAndServe(":"+*middlewarePort, handlers.Router); err != nil {
 		log.Println(err.Error() + " Failed to listen for HTTP")
 	}
 }
--- a/middleware/src/middleware.go
+++ b/middleware/src/middleware.go
@@ -22,7 +22,7 @@ type Middleware struct {
 	info                 rpcmessages.SampleInfoResponse
 	environment          system.Environment
 	events               chan []byte
-	prometheusClient     *prometheus.PromClient
+	prometheusClient     prometheus.Client
 	redisClient          redis.Redis
 	verificationProgress rpcmessages.VerificationProgressResponse
 	isMockmode           bool
@@ -58,7 +58,7 @@ func NewMiddleware(argumentMap map[string]string, mock bool) *Middleware {
 		dummyIsBaseSetup:   false,
 		dummyAdminPassword: "",
 	}
-	middleware.prometheusClient = prometheus.NewPromClient(middleware.environment.GetPrometheusURL())
+	middleware.prometheusClient = prometheus.NewClient(middleware.environment.GetPrometheusURL())
 
 	if !mock {
 		middleware.redisClient = redis.NewClient(middleware.environment.GetRedisPort())
@@ -190,7 +190,7 @@ func (middleware *Middleware) SystemEnv() rpcmessages.GetEnvResponse {
 	return response
 }
 
-// SampleInfo returns the chached SampleInfoResponse struct
+// SampleInfo returns the cached SampleInfoResponse struct
 func (middleware *Middleware) SampleInfo() rpcmessages.SampleInfoResponse {
 	return middleware.info
 }
@@ -401,19 +401,6 @@ func (middleware *Middleware) SetHostname(args rpcmessages.SetHostnameArgs) rpcm
 	return rpcmessages.ErrorResponse{Success: false, Message: "invalid hostname"}
 }
 
-// GetHostname returns a the systems hostname in a GetHostnameResponse
-func (middleware *Middleware) GetHostname() rpcmessages.GetHostnameResponse {
-	// TODO: Implement get hostname from redis
-	// TODO: define error codes for this
-	return rpcmessages.GetHostnameResponse{
-		ErrorResponse: &rpcmessages.ErrorResponse{
-			Success: false,
-			Message: "GetHostname is not implemnted",
-			Code:    rpcmessages.ErrorUnexpected,
-		},
-	}
-}
-
 // EnableTor enables/disables the tor.service and configures bitcoind and lightningd based on the passed ToggleSettingEnable/Disable argument
 // and returns a ErrorResponse indicating if the call was successful.
 func (middleware *Middleware) EnableTor(toggleAction rpcmessages.ToggleSetting) rpcmessages.ErrorResponse {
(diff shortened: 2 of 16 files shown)


middleware/src/middleware.go
```diff
@@ -597,3 +561,99 @@ func (middleware *Middleware) SetRootPassword(args rpcmessages.SetRootPasswordAr
...
+	const (
+		// Redis keys
+		baseHostnameKey      string = "base:hostname"
+		torEnabledKey        string = "tor:base:enabled"
+		middlewareOnionKey   string = "tor:bbbmiddleware:onion"
+		bitcoindListenKey    string = "bitcoind:listen"
+		baseVersionKey       string = "base:version"
+		bitcoindVersioKey    string = "bitcoind:version"
+		lightningdVersionKey string = "lightningd:version"
+		electrsVersionKey    string = "electrs:version"
```
0xB10C: Does it make sense to have those in the `redis` package in e.g. `keys.go`?

0xB10C:
Rebased on top of master and moved the Redis keys to redis/keys.go as commented in https://github.com/digitalbitbox/bitbox-base/pull/218#discussion_r330447544.

middleware/src/middleware_test.go
```diff
@@ -1,15 +1,14 @@
-package middleware_test
+package middleware
```
benma: This is okay, but a few suggestions if you like them (we iterated those guidelines in my former company):

- If you test the public interface of the package without accessing internals, keep `_test`, which does not allow accessing the unexported stuff
- If you do explicitly want to test unexported functions. consider splitting those out into their own package. If unreasonable to do that, name the file `middleware_internal_test.go` and test only internals there, and keep the the public tests in `middleware_test.go`.
- If you only need  to access like one or two little things which are unexported, or you need to set some test mock function which can only be done via internals, you can make a file `middleware_export_test.go` and export your unexported things. See this little tutorial: https://medium.com/@robiplus/golang-trick-export-for-test-aa16cbd7b8cd

The benefit of this is that it's very clear what part of the code is tested. Imho testing only the public interface is best (of course exceptions exist), and testing internals is an indication that the internals are too complicated and need to be split.

middleware/src/prometheus/prometheus.go
```diff
@@ -84,6 +86,43 @@ func (client *PromClient) VerificationProgress() float64 {
...
 	firstResultValue := queryResult[0].Map()["value"].Array()
-	log.Println("result")
 	return firstResultValue[1].Float()
 }
+
+// GetInt queryies prometheus with the provieded query and returns an int64.
+func (client *PromClient) GetInt(query string) (int64, error) {
+	response, err := client.query(query)
+	if err != nil {
+		log.Println(err.Error())
```
benma: `log.Error`?

middleware/src/middleware_test.go
```diff
@@ -1,15 +1,14 @@
...
-	middleware "github.com/digitalbitbox/bitbox-base/middleware/src"
 	"github.com/digitalbitbox/bitbox-base/middleware/src/rpcmessages"
 	"github.com/stretchr/testify/require"
 )
 
 // setupTestMiddleware middleware returns a middleware setup with testing arguments
-func setupTestMiddleware() *middleware.Middleware {
+func setupTestMiddleware() *Middleware {
 	argumentMap := make(map[string]string)
 	argumentMap["bitcoinRPCUser"] = "user"
```
benma: unrelated, but for another PR/issue: this should probably be a struct with well defined fields, otherwise changing the config map probably will not be applied consistently in the future.

middleware/src/rpcserver/rpcserver.go
```diff
@@ -67,7 +66,7 @@ type Middleware interface {
 	ShutdownBase() rpcmessages.ErrorResponse
 	RebootBase() rpcmessages.ErrorResponse
 	EnableRootLogin(rpcmessages.ToggleSetting) rpcmessages.ErrorResponse
-	GetBaseVersion() rpcmessages.GetBaseVersionResponse
+	GetBaseInfo() rpcmessages.GetBaseInfoResponse
```
Tomasvrba: The actual implementation is missing here in `rpcserver.go`

middleware/src/prometheus/prometheus.go
0xB10C: I don't think that exists :) https://golang.org/pkg/log/

middleware/src/middleware_test.go
0xB10C: Thank you for sharing this!

middleware/src/middleware_test.go
0xB10C: Agree. https://github.com/shiftdevices/bitbox-base-internal/issues/299

0xB10C:
Decided to add a commit that does a bit of the log awaited Prometheus refactoring. Otherwise addressed feedback from Tom and Marko.

middleware/src/middleware.go
```diff
@@ -597,3 +561,89 @@ func (middleware *Middleware) SetRootPassword(args rpcmessages.SetRootPasswordAr
 		Code:    rpcmessages.ErrorSetRootPasswordTooShort,
 	}
 }
+
+// GetBaseInfo returns information about the Base in a GetBaseInfoResponse
+func (middleware *Middleware) GetBaseInfo() rpcmessages.GetBaseInfoResponse {
+	const (
+		// Prometheus queries
+		freeDiskspaceQuery  string = "node_filesystem_free_bytes{fstype=\"ext4\", mountpoint=\"/mnt/ssd\"}"
```
benma: `string` redundant here I think

middleware/src/middleware.go
```diff
@@ -597,3 +561,89 @@ func (middleware *Middleware) SetRootPassword(args rpcmessages.SetRootPasswordAr
...
+	}
+
+	electrsVersion, errResp := middleware.getStringFromRedis(redis.ElectrsVersion)
+	if !errResp.Success {
+		return rpcmessages.GetBaseInfoResponse{ErrorResponse: &errResp}
+	}
+
+	return rpcmessages.GetBaseInfoResponse{
+		ErrorResponse: &rpcmessages.ErrorResponse{
+			Success: true,
```
benma: unrelated, but: Success flag looks redundant. ErrorResponse could be `nil` when no error, and then it looks a bit more like Go `if response.Error  != nil { return response.Error } `)

middleware/src/util.go
```diff
@@ -164,3 +167,66 @@ func determineEnableValue(enable rpcmessages.ToggleSetting) string {
...
+	log.Printf("getting '%s' as string from Redis\n", key)
+	value, err := middleware.redisClient.GetString(key)
+	if err != nil {
+		return "", rpcmessages.ErrorResponse{
+			Success: false,
+			Message: fmt.Errorf("could not get %s from redis: %s", key, err.Error()).Error(),
+			Code:    rpcmessages.ErrorRedisError,
+		}
+	}
+	return value, rpcmessages.ErrorResponse{Success: true}
```
benma: if ErrorResponse was `nil` in case of success, this line would also look more natural.

middleware/src/util.go
```diff
@@ -164,3 +167,66 @@ func determineEnableValue(enable rpcmessages.ToggleSetting) string {
 	}
 	return "disable"
 }
+
+// getStringFromRedis returns a string for a given Redis key.
+// If an error occours a ErrorResponse is returned with a ErrorRedisError as error code.
+func (middleware *Middleware) getStringFromRedis(key redis.BaseRedisKey) (value string, errResp rpcmessages.ErrorResponse) {
```
benma: this and the next fewfunctions look like they are about converting the error more so than about the actual query.

A good refactor of this would be to have a function `convertRedisError` and only apply it to the error value.

middleware/src/util.go
```diff
@@ -164,3 +167,66 @@ func determineEnableValue(enable rpcmessages.ToggleSetting) string {
...
+	return value, rpcmessages.ErrorResponse{Success: true}
+}
+
+// getBooleanFromRedis returns a boolean for a given Redis key. This function
+// internally checks if the value for the given key is set to 1.
+// If so, then true is returned, else false.
+// If an error occours a ErrorResponse is returned with a ErrorRedisError as error code.
+func (middleware *Middleware) getBooleanFromRedis(key redis.BaseRedisKey) (value bool, errResp rpcmessages.ErrorResponse) {
+	log.Printf("getting '%s' as boolean from Redis\n", key)
+	val, err := middleware.redisClient.GetInt(key)
```
benma: drop this function, add `GetBoolean` to `redisClient`.

middleware/src/middleware.go
0xB10C: Not addressing this, but I'll keep it in mind.

0xB10C:
Addressed feedback and tested interoperability with  https://github.com/digitalbitbox/bitbox-wallet-app/pull/588

middleware/cmd/middleware/main.go
```diff
@@ -12,6 +12,7 @@ import (
 )
 
 func main() {
+	middlewarePort := flag.String("middlewareport", "8845", "Port the middleware should listen on")
```
TheCharlatan: Can you add the default here?

0xB10C:
Rebased onto master.

PR #224, opened 2019-10-02
Middleware: add /version endpoint 

0xB10C:
The Base needs an unauthenticated endpoint providing the Base version. This Endpoint is provided by the Middleware and reachable under `http://<ip>:8845/version`.

### DIFF
--- a/middleware/src/handlers/handlers.go
+++ b/middleware/src/handlers/handlers.go
@@ -2,6 +2,7 @@
 package handlers
 
 import (
+	"encoding/json"
 	"log"
 	"net/http"
 	"sync"
@@ -16,8 +17,9 @@ import (
 
 // Middleware provides an interface to the middleware package.
 type Middleware interface {
-	// Start triggers the main middleware event loop that emits events to be caught by the handlers.
-	Start() <-chan []byte
+	Start() <-chan []byte // Start triggers the main middleware event loop that emits events to be caught by the handlers.
+
+	/* --- RPCs --- */
 	SystemEnv() rpcmessages.GetEnvResponse
 	SampleInfo() rpcmessages.SampleInfoResponse
 	ResyncBitcoin() rpcmessages.ErrorResponse
@@ -41,6 +43,9 @@ type Middleware interface {
 	VerificationProgress() rpcmessages.VerificationProgressResponse
 	UserAuthenticate(rpcmessages.UserAuthenticateArgs) rpcmessages.ErrorResponse
 	UserChangePassword(rpcmessages.UserChangePasswordArgs) rpcmessages.ErrorResponse
+	/* --- RPCs end --- */
+
+	BaseVersionForVersionEndpoint() string
 }
 
 // Handlers provides a web api
@@ -70,6 +75,7 @@ func NewHandlers(middlewareInstance Middleware, dataDir string) *Handlers {
 		clientsMap:  make(map[int]chan<- []byte),
 	}
 	handlers.Router.HandleFunc("/", handlers.rootHandler).Methods("GET")
+	handlers.Router.HandleFunc("/version", handlers.versionHandler).Methods("GET")
 	handlers.Router.HandleFunc("/ws", handlers.wsHandler)
 
 	handlers.middlewareEvents = handlers.middleware.Start()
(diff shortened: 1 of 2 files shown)


0xB10C:
Fixed liter error

middleware/src/middleware.go
```diff
@@ -597,3 +599,38 @@ func (middleware *Middleware) SetRootPassword(args rpcmessages.SetRootPasswordAr
...
+// The version is primarily obtained via a Redis GET.
+// If Redis can not be reached, the version is read from disk.
+// If both redis and the disk can not supply a version, then "0.0.0" is returned.
+// **This is not a RPC endpoint function.**
+func (middleware *Middleware) BaseVersionForVersionEndpoint() string {
+	//const versionFilePath string = "/opt/shift/config/version"
+	const versionFilePath string = "/tmp/version"
+	const fallbackVersion string = "0.0.0"
+
+	version, err := middleware.redisClient.GetString("base:version") // TODO: have this as a const somewhere
```
0xB10C: `TODO` could be resolved with something as discussed here: https://github.com/digitalbitbox/bitbox-base/pull/218#discussion_r330447544

middleware/src/handlers/handlers.go
```diff
@@ -103,6 +109,30 @@ func (handlers *Handlers) rootHandler(w http.ResponseWriter, r *http.Request) {
 	}
 }
 
+// versionHandler provides an endpoint supplying the Base version as JSON
+// This endpoint is curitial and needed in every Base release.
```
TheCharlatan: critical?

middleware/src/handlers/handlers.go
```diff
@@ -103,6 +109,30 @@ func (handlers *Handlers) rootHandler(w http.ResponseWriter, r *http.Request) {
...
+// It does not have to be implemented in the Middleware, but
```
TheCharlatan: , but it definitely needs to be in Candy Mountain.

middleware/src/middleware.go
TheCharlatan: Maybe I am lacking some context here, but why should be const? It's typically only constant at runtime, but not at compile time.

middleware/src/handlers/handlers.go
```diff
@@ -41,6 +43,9 @@ type Middleware interface {
 	VerificationProgress() rpcmessages.VerificationProgressResponse
 	UserAuthenticate(rpcmessages.UserAuthenticateArgs) rpcmessages.ErrorResponse
 	UserChangePassword(rpcmessages.UserChangePasswordArgs) rpcmessages.ErrorResponse
+	/* --- RPCs end --- */
```
TheCharlatan: Nice :)

middleware/src/middleware.go
0xB10C: Oh sorry, that TODO is not clear. 

I thought about having the Redis keys (e.g. `base:version`) as a const in something like `redis/keys.go` in https://github.com/digitalbitbox/bitbox-base/pull/218#discussion_r330447544

TheCharlatan: Ah, I see.

0xB10C:
A user won't be able to update the Middleware separately, so the Base version is used to identify what protocol version to use.

> From the version endpoint of the middleware I would expect the middleware version...

Yep, somewhat agree. Yet the Middleware is used because:

> Why middleware and not nginx static files?
> - Better in one place than scattered in many
> - partly dynamic
> - might add more dynamic endpoints in the future

See prior discussion https://github.com/shiftdevices/bitbox-base-internal/issues/275#issuecomment-535086191

TheCharlatan:
While we currently have no intention to ship the middleware separately, this might change in the future. A user could also get access to the base and downgrade the middleware manually. 
It also feels very weird to me to have the base version as a slave of the middleware version. When we make protocol changes in the middleware now, how the version for the base will be bumped. Also changes to the base will now produce version changes that have nothing to do with changes in the middleware's protocol. I don't think centralizing this is a good idea. All the services running on the base have their own version and the middleware should have one as well in my opinion.
If I understand @benma's point correctly the important thing is the middleware version, since it dictates the protocol, and not the base version.

0xB10C:
Agree to expose a middleware version. I'd tend to say that having a both, a middleware version (for the protocol communication) and a Base version, could be beneficial in the future. Yet I'm not sure in which situation exactly. What do you think @TheCharlatan @benma?

TheCharlatan:
I think the base version would fit nicely into the GetBaseInfo rpc.

Stadicus:
Agreed to have middleware version in unauthenticated api /version and base version in GetBaseInfo rpc.

0xB10C:
Closing in favor of https://github.com/digitalbitbox/bitbox-base/pull/225 which implements the middleware version.

PR #228, opened 2019-10-15
Middleware: add Base update RPCs

0xB10C:
The BitBoxBase should support over-the-air firmware updates. This commit adds the two necessary RPCs for this.

**UpdateBase** RPC:
This RPC performs an update of the Base firmware to a specified version. This includes downloading, applying and rebooting. Every time the download progress or the state changes a notification is send to the App.

**GetBaseUpdateProgress** RPC:
This RPC should be called when the app receives a notification about a changed state. It returns the state, downloaded byte and progress in percent.

This implements step 6, 8 and 10 from https://github.com/shiftdevices/bitbox-base-internal/issues/266.

Implementation on the App side: https://github.com/digitalbitbox/bitbox-wallet-app/pull/605


middleware/src/middleware.go
```diff
@@ -556,6 +563,154 @@ func (middleware *Middleware) RebootBase() rpcmessages.ErrorResponse {
...
+				middleware.baseUpdateProgress.ProgressDownloadedKiB = downloadedKiB
+				if percentage < 98 {
+					middleware.setBaseUpdateStateAndNotify(rpcmessages.UpdateDownloading)
+				} else {
+					// switch State from `UpdateDownloading` to `UpdateApplying`
+					// This is done at 98% or above since the mender-install script does
+					// only log 100% after applying the update.
+					middleware.setBaseUpdateStateAndNotify(rpcmessages.UpdateApplying)
+				}
+			}
```
TheCharlatan: How does this for loop break in the normal case?

0xB10C: Once EOF is read from stdout. https://github.com/digitalbitbox/bitbox-base/pull/228/files/79f263847831dd1862a21fd202ddec22be28544b#diff-bb48a3164edaae7002e0a4949d26fd9dR683-R685

Will add a comment to the for-loop clarifying that.

0xB10C:
rebased on master

PR #231, opened 2019-10-16
Middleware: add first version of GetServiceInfo RPC

0xB10C:
This commit adds a first version of the GetServiceInfo RPC.

The RPC returns information about services such as bitcoind, electrs and lightningd. Information includes for example the current block height or the count of connected peers.

The App should call this RPC once it receives a notification that the serviceInfo in the middleware has changed.

This RPC will likely be extended with additional information in the future.

### DIFF
--- a/middleware/src/middleware.go
+++ b/middleware/src/middleware.go
@@ -17,7 +17,7 @@ import (
 	lightning "github.com/fiatjaf/lightningd-gjson-rpc"
 )
 
-// Middleware connects to services on the base with provided parrameters and emits events for the handler.
+// Middleware connects to services on the base with provided parameters and emits events for the handler.
 type Middleware struct {
 	info                 rpcmessages.SampleInfoResponse
 	environment          system.Environment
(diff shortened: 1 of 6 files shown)


middleware/src/middleware.go
```diff
@@ -136,6 +137,9 @@ func (middleware *Middleware) GetVerificationProgress() bool {
 // rpcLoop gets new data from the various rpc connections of the middleware and emits events if new data is available
 func (middleware *Middleware) rpcLoop() {
 	for {
+		if middleware.didServiceInfoChange() {
+			middleware.events <- []byte(rpcmessages.OpServiceInfoChanged)
+		}
 		if middleware.GetSampleInfo() {
```
TheCharlatan: Lets finally retire the last of the sample/demo calls. Just need to make sure that they are properly removed from the app frontend as well.

0xB10C: Yeah. Will do that in a separate PR as discussed with Stadicus a few days ago. Enables us to merge whenever the App stops using the old demo calls.

0xB10C:
rebased on master

0xB10C:
... and removed a leading new line

PR #235, opened 2019-10-28
Middleware: add Base firmware update check and notification

0xB10C:
This adds a periodic check for firmware updates to the middleware.
If an update is available, a the middleware sends a notification.
The App can get more information about the update by calling the
IsBaseUpdateAvailable RPC, which returns a boolean indicating if
an update is available and information about the update.

The App should call `IsBaseUpdateAvailable` after the connection is
established.

### DIFF
--- a/middleware/cmd/middleware/main.go
+++ b/middleware/cmd/middleware/main.go
@@ -29,6 +29,7 @@ func main() {
 	prometheusURL := flag.String("prometheusurl", "http://localhost:9090", "Url of the prometheus server in the form of 'http://localhost:9090'")
 	redisPort := flag.String("redisport", "6379", "Port of the Redis server")
 	redisMock := flag.Bool("redismock", false, "Mock redis for development instead of connecting to a redis server, default is 'false', use 'true' as an argument to mock")
+	imageUpdateInfoURL := flag.String("updateinfourl", "https://shiftcrypto.ch/updates/base.json", "URL to query information about updates from (defaults to https://shiftcrypto.ch/updates/base.json)")
 	flag.Parse()
 
 	argumentMap := make(map[string]string)
@@ -43,6 +44,7 @@ func main() {
 	argumentMap["bbbCmdScript"] = *bbbCmdScript
 	argumentMap["prometheusURL"] = *prometheusURL
 	argumentMap["redisPort"] = *redisPort
+	argumentMap["imageUpdateInfoURL"] = *imageUpdateInfoURL
 	argumentMap["middlewareVersion"] = version
 
 	logBeforeExit := func() {
--- a/middleware/src/middleware.go
+++ b/middleware/src/middleware.go
@@ -17,6 +17,7 @@ import (
 	"github.com/digitalbitbox/bitbox-base/middleware/src/redis"
 	"github.com/digitalbitbox/bitbox-base/middleware/src/rpcmessages"
 	"github.com/digitalbitbox/bitbox-base/middleware/src/system"
+	"github.com/digitalbitbox/bitbox-base/middleware/src/util/semver"
 	lightning "github.com/fiatjaf/lightningd-gjson-rpc"
 	"golang.org/x/crypto/bcrypt"
 )
@@ -41,6 +42,8 @@ type Middleware struct {
 	verificationProgress rpcmessages.VerificationProgressResponse
 	serviceInfo          rpcmessages.GetServiceInfoResponse
 	baseUpdateProgress   rpcmessages.GetBaseUpdateProgressResponse
+	baseUpdateAvailable  rpcmessages.IsBaseUpdateAvailableResponse
+	baseVersion          *semver.SemVer
 	isMockmode           bool
 	// Saves state for the setup process
 	isMiddlewarePasswordSet bool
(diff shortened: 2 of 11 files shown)


middleware/cmd/middleware/main.go
```diff
@@ -42,6 +43,7 @@ func main() {
 	argumentMap["bbbCmdScript"] = *bbbCmdScript
 	argumentMap["prometheusURL"] = *prometheusURL
 	argumentMap["redisPort"] = *redisPort
+	argumentMap["firmwareUpdateInfoURL"] = "https://gist.githubusercontent.com/0xB10C/a0e067b7a6e6fdcbdce2fb1f172ec7a5/raw/85136861daa70235e4a843931b123824bb6a69e5/base-icecream.json" //*firmwareUpdateInfoURL
```
0xB10C: For review and testing. Will change back to 

```suggestion
	argumentMap["firmwareUpdateInfoURL"] = firmwareUpdateInfoURL
```

before merge.

middleware/src/middleware.go
```diff
@@ -159,9 +173,82 @@ func (middleware *Middleware) rpcLoop() {
 	}
 }
 
+// updateCheckLoop repeatedly checks for information about new Base firmware updates
+// When an update is available it's
+func (middleware *Middleware) updateCheckLoop() {
+	// This time is chosen arbitrary, but the time should be not too high for users to be notified
+	// not to long after the update release and not too low to avoid to frequent update checks.
+	const timeBetweenUpdateChecks time.Duration = 30 * time.Minute
```
0xB10C: I personally think a 30 minute update check interval is a good compromise between too frequent and to infrequent. But this is chosen arbitrarily. Looking for other suggestions.

middleware/src/middleware.go
```diff
@@ -159,9 +173,82 @@ func (middleware *Middleware) rpcLoop() {
...
+		}
+
+		time.Sleep(timeBetweenUpdateChecks)
+	}
+}
+
+// getBaseUpdateInfo GETs a JSON file over HTTP which includes information about the update.
+func getBaseUpdateInfo(url string) (updateInfo rpcmessages.UpdateInfo, err error) {
+	client := http.Client{
+		Timeout: 15 * time.Second, // timeout is chosen arbitrary, but should account for (very) slow tor connections.
```
0xB10C: The HTTP connection timeout of 15 seconds is chosen arbitrary. It should account for very slow tor connections. Let me know what you think.

TheCharlatan:
Concept Nit: Since the middleware will also have to handle the updates of the HSM firmware, I'd propose calling it 'Base Image' instead of 'Base Firmware'

middleware/src/util/semver/semver.go
```diff
@@ -0,0 +1,110 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// This file is copied from the bitbox-wallet-app (https://github.com/digitalbitbox/bitbox-wallet-app/blob/master/util/semver/semver.go).
+// Modifications:
+// - change errp to errors.New()
```
0xB10C: I copy and pasted the semver.go packet from the App. I'm not to happy with it, but I think it's better to `vendor/` bigger parts of the App into the Base repository. Happy for suggestions here as well (especially from @benma).

middleware/src/util/semver/semver.go
```diff
@@ -0,0 +1,110 @@
+// Copyright 2018 Shift Devices AG
```
benma: could update this to, 2019 Shift Cryptosecurity AG

middleware/cmd/middleware/main.go
```diff
@@ -42,6 +43,7 @@ func main() {
 	argumentMap["bbbCmdScript"] = *bbbCmdScript
 	argumentMap["prometheusURL"] = *prometheusURL
 	argumentMap["redisPort"] = *redisPort
+	argumentMap["imageUpdateInfoURL"] = "https://gist.githubusercontent.com/0xB10C/a0e067b7a6e6fdcbdce2fb1f172ec7a5/raw/85136861daa70235e4a843931b123824bb6a69e5/base-icecream.json" //*imageUpdateInfoURL
```
benma: a gist? let's use shiftcrypto.ch like for the App, or maybe github.com/digitalbitbox/bitbox-base/releases/?

middleware/src/middleware.go
Stadicus: I think twice a day would suffice as well, if we could trigger a manual check from the App (e.g. if we announce an update), but that's not really important. I'm fine with just checking every 30 minutes.

Stadicus:
> this should be changed on a repository wide (e.g. docs & shell scripts) level (@Stadicus).

Addressed in https://github.com/digitalbitbox/bitbox-base/pull/236

middleware/src/middleware.go
0xB10C: Will leave it at 30 minutes.

middleware/src/middleware.go
0xB10C: Going with 15 seconds here.

middleware/cmd/middleware/main.go
0xB10C: Sure. The gist is for testing and review as described above.

0xB10C:
Opend a PR for the `base.json` file on shiftcrypto.ch https://github.com/shiftdevices/website_design/pull/210

0xB10C:
Rebased on to #227. https://github.com/digitalbitbox/bitbox-base/pull/235/commits/37551f98226bc90f44a1c43e603e9c9cbfcd1b50 adds the authentication stuff. @TheCharlatan please have a look if I missed anything. 
Maybe https://github.com/shiftdevices/bitbox-base-internal/issues/269 could need an update as well.

TheCharlatan:
lgtm

0xB10C:
Rebased and resolved merge conflicts.

PR #238, opened 2019-10-29
Add factory reset framework and implement authentication reset

Stadicus:
Implements shiftdevices/bitbox-base-internal#308

**Add 'reset' features for authentication**
* Users need to be able to reset various aspects of the Base config.
* With `bbb-cmd.sh reset <auth|config|image|ssd>` different components can be reset to factory settings.
* When run manually, user needs to confirm the reset by typing `YES`. For non-interactive usage, the flag `--assume-yes` can be used.

The commit
* adds the overall framework for all 'reset' commands, without adding the specific functionality.
* Implements the first reset command `bbb-cmd.sh reset auth`

**Check reset trigger / token on flashdrive when booting**
* User must be able to reset a forgotten password.
* The App is of not much use, because the user can no longer access the Base using the App.
* On boot, the Base checks if a flashdrive is present and looks for certain trigger files (e.g. `reset-base-auth`)
* As a password reset gives full access to the Base, the flashdrive must contain a reset token that is created on backup.
* The hashed reset token is compared with the hash that is stored on the Base and if they match, the reset actions are triggered.

The commit
* extends `bbb-cmd.sh backup sysconfig` to also store a restore token on the flashdrive, and keep the hashed token on the Base
* mounts flashdrive on boot in `systemd-update-checks.sh`, checks if a valid reset token is present and triggers reset actions

### DIFF
--- a/armbian/base/scripts/bbb-cmd.sh
+++ b/armbian/base/scripts/bbb-cmd.sh
@@ -242,6 +242,9 @@ case "${MODULE}" in
                     echo "FLASHDRIVE MOUNT: device ${ARG} is not unique and/or has partitions."
                     errorExit FLASHDRIVE_MOUNT_NOT_UNIQUE
 
+                elif mountpoint /mnt/backup -q; then
+                    echo "FLASHDRIVE MOUNT: mountpoint /mnt/backup is already in use. Assuming prior mount, no error."
+
                 else
                     scsidev=$(lsblk -o NAME,SIZE,FSTYPE -abrnp -I 8 "${ARG}")
                     name=$( echo "${scsidev}" | cut -s -f 1 -d " " )
@@ -306,6 +309,19 @@ case "${MODULE}" in
                     errorExit BACKUP_SYSCONFIG_NOT_A_MOUNTPOINT
                 fi
                 echo "OK: backup created as /mnt/backup/bbb-backup.rdb"
+
+                # add Factory Reset token
+                RESET_TOKEN="$(< /dev/urandom tr -dc A-Za-z0-9 | head -c64)"
+                RESET_TOKEN_HASH=$(echo -n "${RESET_TOKEN}" | sha256sum | tr -d "[:space:]-")
+
+                # write reset token to usb drive, no linebreak allowed
+                printf "%s" "${RESET_TOKEN}" > /mnt/backup/.reset-token
+
+                # append reset token hash for permission check locally
+                echo "${RESET_TOKEN_HASH}" >> /data/reset-token-hashes
+                chmod 600 /data/reset-token-hashes
+                echo "OK: reset token created on flashdrive"
+
                 ;;
 
             # backup c-lightning on-chain keys in 'hsm_secret' into Redis database
--- a/armbian/base/scripts/bbb-cmd.sh
+++ b/armbian/base/scripts/bbb-cmd.sh
@@ -16,6 +16,7 @@ possible commands:
   flashdrive    <check|mount|umount>
   backup        <sysconfig|hsm_secret>
   restore       <sysconfig|hsm_secret>
+  reset         <auth|config|image|ssd>
   mender-update <install|commit>
 
 "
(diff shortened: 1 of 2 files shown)


Stadicus:
Example output when running the script `systemd-update-check.sh` with a flashdrive plugged in:

```
root@bitbox-base:/mnt/backup# ls -la
drwxr-xr-x 3 root root     16384 Jan  1  1970 .
drwxr-xr-x 1 root root        60 Oct 29 21:10 ..
-rwxr-xr-x 1 root root        65 Oct 29 21:29 .reset-token
-rwxr-xr-x 1 root root      2587 Oct 29 21:29 bbb-backup.rdb
-rwxr-xr-x 1 root root         0 Oct 29 21:57 reset-base-auth
```
```
root@bitbox-base:~# /opt/shift/scripts/systemd-update-checks.sh
...
RESET: flashdrive mounted
RESET: reset token present on flashdrive, hashed value: 9bbda7cad7cffc268ae424204459a8f823c8a3ab2352ff6fbba884e5cdc11461
RESET: valid reset token found
RESET: trigger file 'reset-base-auth' found, initiating reset of authentication
INFO: reset confirmed with '--assume-yes'
INFO: set Redis key 'base:setup' to '0'
INFO: set Redis key 'middleware:passwordSetup' to '0'
OK: middleware authentication reset, setup wizard can be run again.
...
```

Stadicus:
Rebased to master

Stadicus:
Rebased to master

0xB10C:
Haven't looked at the code yet. This is concept review.

To clarify the user-perspective a bit what needs to be done for the user to reset his password (from what I understood):

1. User gets his USB stick with the Backup and uses an external computer, e.g. Laptop, to create a reset-trigger file.
2. Plugs USB stick into Base
3. Force restarts Base via physical buttons
4. Base detects the USB device and resets user auth (hopefully deletes the trigger file)
5. User can redo the setup wizard. 


Things to keep in mind here:
- Step 1: Backup gets into contact with possible malware infected computer.
- Step 4: should delete the trigger file on success, otherwise the Base would be reset on every reboot, if the USB stick is not removed.
- Step 5: the newly written reset-token should be different from the pre-reset reset-token.

In general creating trigger files on USB sticks and rebooting the device is bad UX wise.

---


> The App is of not much use, because the user can no longer access the Base using the App.

I disagree. As long as the user only forgot the password and did not mess up the system he probably can. Noise channel setup should work without authentication. What the reason to not use a unauthenticated, but Noise encrypted, `ResetAuthetication` RPC?

1. User clicks on "forgot password" under the normal login screen.
2. App prompts user to plug backup USB stick in.
3. App calls `ResetAuthetication` RPC
3.1. Mounts USB
3.2. Checks secret
3.3. Resets Auth
3.4 Unmounts USB stick
4. App shows the User the wizard or just a password set screen.

Why wouldn't that work?

Stadicus:
> In general creating trigger files on USB sticks and rebooting the device is bad UX wise.

I agree, this is meant as 
* a general implementation of the reset functionality in `bbb-cmd.sh` not depending on the trigger mechanism, and 
* a first, very low-level trigger mechanism for the authentication reset

We can add additional trigger mechanisms (e.g. from the HSM, or the App) as well, and we should. 

>> The App is of not much use, because the user can no longer access the Base using the App.

> I disagree.

You're right for the authentication reset, an unauthenticated RPC would also work. But especially with the 'config' reset (fixing a broken configuration, maybe preventing the Middleware from running), and the 'image' reset (e.g. flashing a firmware from the eMMC or also present on the flashdrive) I think it makes sense to have these triggers from USB as a measure of last resort.

Stadicus:
> Step 4: should delete the trigger file on success, otherwise the Base would be reset on every reboot, if the USB stick is not removed.

Yes, the trigger file is renamed, e.g. to `reset-base-auth.done`

Stadicus:
> Step 5: the newly written reset-token should be different from the pre-reset reset-token.

A reset does not necessarily create a new token. For now, all created tokens stay valid and their hashes are appended to the `/data/reset-token-hashes` file. This file could be deleted to invalidate all previous tokens, but that is not implemented yet.

armbian/base/scripts/systemd-update-checks.sh
```diff
@@ -37,6 +37,54 @@ redis_set "base:version" "${VERSION}"
...
 
+# check for reset triggers on flashdrive
+if FLASHDRIVE="$(/opt/shift/scripts/bbb-cmd.sh flashdrive check)"; then
+    echo "RESET: device ${FLASHDRIVE} detected"
+    if /opt/shift/scripts/bbb-cmd.sh flashdrive mount "${FLASHDRIVE}"; then
+        echo "RESET: flashdrive mounted"
+
+        # are all necessary files for a reset present?
+        if [[ -f /mnt/backup/.reset-token ]] && [[ -f /data/reset-token-hashes ]]; then
+            FLASHDRIVE_TOKEN_HASH="$(echo -n "$(</mnt/backup/.reset-token)" | sha256sum | tr -d "[:space:]-")"
```
TheCharlatan: This seems really unsafe. What if the file contains another expression, for example `$($(ls)\n`? This expression will first be interpreted by the 256sum and could be used to point to the /data/reset-token-hashes files. Maybe run the read string through a sanitizer first?

TheCharlatan:
I like the concept a lot, but lets just take the time to review this carefully, since it is quite sensitive and bash is a giant can of worms when it comes to sensitive user input.

armbian/base/scripts/systemd-update-checks.sh
Stadicus: Good input thanks. I can easiyl sanatize the token, as its a SHA256 string with only hex characters allowed.

Example
```console
$ RESET_TOKEN='abcdefghijklmnopqrstuvwxyzABCDXYZ1234567890!@#$%^&*()_+/?|}{}|":' && echo $RESET_TOKEN && echo ${RESET_TOKEN//[^a-f0-9]/}
abcdefghijklmnopqrstuvwxyzABCDXYZ1234567890!@#$%^&*()_+/?|}{}|":
abcdef1234567890
```

Change to above script: sanitize the `FLASHDRIVE_TOKEN` first before using it to create the `FLASHDRIVE_TOKEN_HASH`:

```bash
FLASHDRIVE_TOKEN=$(head -n1 /mnt/backup/.reset-token)
FLASHDRIVE_TOKEN=${FLASHDRIVE_TOKEN//[^a-f0-9]/}
FLASHDRIVE_TOKEN_HASH="$(echo -n "${FLASHDRIVE_TOKEN}" | sha256sum | tr -d "[:space:]-")"
```

What do you think?

Stadicus: Also discussed with @NickeZ to validate that this approach is not prone to code injection.

TheCharlatan: Why not just sha256sum the file, e.g. `FLASHDRIVE_TOKEN_HASH=sha256sum /mnt/backup/.reset-token`?

Stadicus: I changed the creation of the reset token so that it does not contain anything else, not even a line break. That makes the hash of the whole file identical to the string itself, and there is no need to sanitize the reset token content.

Most elegant solution yet:
```
FLASHDRIVE_TOKEN_HASH="$(sha256sum /mnt/backup/.reset-token | cut -f 1 -d " ")"
```

armbian/base/scripts/bbb-cmd.sh
```diff
@@ -435,6 +452,48 @@ case "${MODULE}" in
...
+            # reset authentication for running BitBoxApp setup wizard again
+            AUTH)
+                redis_set "base:setup" 0
+                redis_set "middleware:passwordSetup" 0
+
+                echo "OK: middleware authentication reset, setup wizard can be run again."
+                ;;
+
+            *)
+                echo "Invalid argument for module ${MODULE}: command ${COMMAND} unknown."
```
0xB10C: nit: could add a not implemented notice via an `echo` for the the `config`, `image` and `ssd` commands. Not sure if it's worth doing it.

Stadicus: As these commands are not set in stone yet, I think the script just failing is fine. Need to add proper error messages once we think about additional functionality and know what commands will be supported in the future.

PR #241, opened 2019-11-04
update 'build-ondevice' make target

Stadicus:
This pull request updates the build method to run the configuration script directly on a preinstalled Linux distribution. It's used mainly for testing of non-Armbian distributions.

**compile Go: use build instead of install in Docker**
https://github.com/shiftdevices/bitbox-base-internal/issues/326

Because:
* When cross-compiling the Go tools inside Docker, with the command
  'docker install', the binaries are put into the directory
  $GOPATH/bin/linux_arm64
* One goal is to run installation (incl. Go compilation) directly on
  the ARM device, to support other distributions. When not cross-
  compiling, the binaries are put into $GOPATH/bin/, failing the
  make process.

This commit:
* uses 'go build' instead of 'go install' to always put the binaries
  directly in the source directory and copy them from there.

**Update 'make build-ondevice'**
Because:
* Users should be able to run the BitBoxBase customization on a vanilla
  Linux installation, e.g. the Ayufan ARM distribution.
* For that to work, Go binaries need to be compiled first.
* Armbian specific changes are allowed to be skipped in that build mode

This commit:
* checks if Go tool binaries are available, or aborts
* adds some checks in customization script when running on device
* in prometheus-base.sh, check if Redis values like 'build:date' or
  'build:commit' are null and return 'n/a'.

### DIFF
--- a/armbian/armbian-build.sh
+++ b/armbian/armbian-build.sh
@@ -65,10 +65,19 @@ case ${ACTION} in
 		;;
 
 	ondevice)
-    	# copy custom scripts to filesystem
-    	mkdir -p /opt/shift
-    	cp -aR base/* /opt/shift
-    	chmod -R +x /opt/shift/scripts
+		# copy custom scripts to filesystem
+		mkdir -p /opt/shift
+		cp -aR base/* /opt/shift
+		chmod -R +x /opt/shift/scripts
+
+		# check dependency: Go binaries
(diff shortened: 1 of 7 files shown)


Stadicus:
Rebased to master after changes by @benma to Makefiles in https://github.com/digitalbitbox/bitbox-base/commit/94cb2f302a1601313b826e1a4d53204cdfd822f1#diff-f345a466d51d2c7dd83b95180dd44c48

Stadicus:
I'll continue to expand this pull request and submit it again.

PR #245, opened 2019-11-06
typo: fix Avaliable > Available for BaseUpdates

Tomasvrba:


Tomasvrba:
> At least my typo was consistent...

It was (half)-correct in rpcmessages :D

PR #246, opened 2019-11-07
prometheus: prevent bitcoind zero values

Stadicus:
https://github.com/shiftdevices/bitbox-base-internal/issues/334

Because:
* The bitcoind scraper (prometheus-bitcoind.py) starts the Prometheus http server immediately, even if bitcoind is not ready yet.
* Metrics are then displayed as zero values, which are stored into the Prometheus time-series database.
* This leads to sudden drops in blockheight, verification process to zero on every reboot.

This commit:
* makes sure that the http server is only started after bitcoind RPCs are available

### DIFF
--- a/armbian/base/scripts/prometheus-bitcoind.py
+++ b/armbian/base/scripts/prometheus-bitcoind.py
@@ -121,19 +121,24 @@ def get_raw_tx(txid):
 
 
 def main():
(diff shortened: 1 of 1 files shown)


0xB10C:
^Code Review ACK

PR #247, opened 2019-11-07
Replace 'BitBox Base' with  'BitBoxBase'

Stadicus:
Because:
* In the past, the project/product was called 'BitBox Base' and that is reflected accross documentation, scripts and code comments.
* It's better to consistently change all occurrences now to the new naming of 'BitBoxBase' instead of doing it gradually.

This commit:
* replaces case-sensitive 'BitBox Base' with 'BitBoxBase'
* no working code is affected, only docs, code comments and one err msg


0xB10C:
Readme logo is old as well. Could be addressed here.

PR #248, opened 2019-11-07
docs: add 'Tinkering' section

Stadicus:
Beta testers working with the device should have a good first explanation of the steps we are using in development regularly, but could be hard to figure out without guidance.

* adds a 'Tinkering' section to the technical documentation
* extends the 'Helper scripts' docs with additional details for
  'bbb-config.sh'
* fixes some broken links


Stadicus:
I'll extend this pull-request to contain more references and also update the rest of the documentation.

PR #249, opened 2019-11-07
add: read IPC notifications from a named pipe

0xB10C:
This defines IPC notifications and adds a IPC notification reader,
which creates and opens a named pipe. Other processes, for example
shell or python scripts, can write notifications into the named pipe.

Notifications are sent in a JSON encoded string. The protocol in
version 1 requires notifications to have a version, topic and a
payload. Notifications can have a maximum length of 4096-1 byte.

Can be tested with:
```shell
echo '{"version": 1, "topic": "sampletopic", "payload": {"testInt":123,"testString": "string", "testBool": true}}' >> /tmp/middleware-notification.pipe
```

### DIFF
--- a/middleware/src/ipcnotification/ipcnotification.go
+++ b/middleware/src/ipcnotification/ipcnotification.go
@@ -1,0 +1,136 @@
+// Package ipcnotification reads notifications from a named pipe and passes them
+// into a channel.
+package ipcnotification
+
+import (
+	"bufio"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"log"
+	"os"
+	"syscall"
+)
+
+// Notification represents an IPC notification that is passed via a named pipe.
+type Notification struct {
+	Version int         `json:"version"`
+	Topic   string      `json:"topic"`
+	Payload interface{} `json:"payload"`
+}
+
+func (notification *Notification) String() string {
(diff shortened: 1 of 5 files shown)


middleware/src/system/system.go
```diff
@@ -94,3 +96,8 @@ func (environment *Environment) GetMiddlewarePort() string {
 func (environment *Environment) GetImageUpdateInfoURL() string {
 	return environment.imageUpdateInfoURL
 }
+
+// GetNotificationNamedPipePath is a getter for the path where the middleware creates and looks for the
+func (environment *Environment) GetNotificationNamedPipePath() string {
```
0xB10C: Test for this is missing.

middleware/src/ipcnotification/ipcnotification.go
```diff
@@ -0,0 +1,131 @@
...
+	Notifications chan Notification
+	closing       bool
+	filePath      string
+	namedPipe     *os.File
+}
+
+// NewReader returns a new Reader that starts reading from the named pipe passed.
+func NewReader(filepath string) (*Reader, error) {
+	reader := &Reader{
+		Notifications: make(chan Notification),
```
TheCharlatan: I would make this private and expose it through a getter, but that is just my personal taste.

0xB10C: I agree with making it private. No need to be potentially reassigned by somebody else.

PR #250, opened 2019-11-07
enable Bitcoin services only after setup

Stadicus:
Currently, all services are enabled and as soon as the BitBoxBase has power and network it will start syncing over Tor.

As we give the user different options during the Setup Wizard, services should not start before that.

This pull-request...
* adds the build option BASE_ENABLE_BITCOIN_SERVICES (default: false)
* does not enable bitcoind, lightningd and electrs systemd units during the build process
* introduces the Redis key base:bitcoind-services:enabled
* adds the command 'bbb-config.sh enable/disable bitcoin-services' to configure systemd and set the Redis key
* script 'systemd-update-check.sh' checks the Redis key on boot and enables the services if -eq 1 (the are not disabled otherwise)

### DIFF
--- a/armbian/base/scripts/bbb-config.sh
+++ b/armbian/base/scripts/bbb-config.sh
@@ -14,7 +14,7 @@ usage: bbb-config.sh [--version] [--help]
 assumes Redis database running to be used with 'redis-cli'
 
 possible commands:
-  enable    <bitcoin_incoming|bitcoin_ibd|bitcoin_ibd_clearnet|dashboard_hdmi|
+  enable    <bitcoin_services|bitcoin_incoming|bitcoin_ibd|bitcoin_ibd_clearnet|dashboard_hdmi|
              dashboard_web|wifi|autosetup_ssd|tor|tor_bbbmiddleware|tor_ssh|
              tor_electrum|overlayroot|sshpwlogin|rootlogin|unsigned_updates>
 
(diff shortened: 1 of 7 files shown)


Stadicus:
Works as expected. On regular update, services are enabled and then started as part of the system checks. On initial flash, services are disabled and must be enabled with `bbb-config.sh enable bitcoin_services`.

After completing the Setup Wizard, the Middleware must call this script. But then it must also start the services by calling `systemd start` for each of them. Is that ok, or should this also be encapsulated into a script call? @0xB10C

0xB10C:
I think it would be cleaner to have one script that is called from the middleware's `FinalizeSetupWizard` RPC call (https://github.com/shiftdevices/bitbox-base-internal/issues/312). 


Procedure for the `FinalizeSetupWizard` would then be:
1. set redis key `base:setup`
2. `bbb-config.sh enable bitcoin-services`
3. `bbb-config.sh start bitcoin-services` (or similar)

correct?

Stadicus:
As there currently is no `start` command in `bbb-config.sh` and we already have a helper script for systemd services, I'll add the functionality to start the bitcoin-related services here:

```
bbb-systemctl.sh start-bitcoin-services
```

0xB10C:
```
bbb-systemctl.sh start-bitcoin-services
```

The script does not / can not return an error, when the services can not be started, correct?

Stadicus:
> The script does not / can not return an error, when the services can not be started, correct?

If systemctl itself fails, the script aborts with exit code 1 (or whatever systemctl returns), but no custom error code.

0xB10C:
Is it possible to add an error code?

Stadicus:
> Is it possible to add an error code?

I could add custom error handling, e.g. like this. Would likely also enhance the script with MockMode.
```bash
start-bitcoin-services)
    checkMockMode

    # use custom error handling
    set +e

    if ! systemctl is-active -q bitcoind.service; then
        systemctl start bitcoind.service
        if [[ ${?} -ne 0 ]]; then
            errorExit SYSTEMD_SERVICESTART_FAILED
        fi
        echo "OK: bitcoind.service started"

    else
        echo "INFO: bitcoind.service already running"
    fi
```

0xB10C:
Will integrate the `SYSTEMD_SERVICESTART_FAILED` into the middleware.

Stadicus:
Added custom error handling and MockMode

0xB10C:
Sounds good @Tomasvrba. I somehow forgot that `systemctl restart` will
start the services even if they are `disabled`.

Stadicus:
> The solution would be to check if base:setup 0 in all of these calls and immediately restart the service only if the setup 1

Added these checks in a separate commit https://github.com/digitalbitbox/bitbox-base/pull/250/commits/0981d2867fa5a02f3e342460e05aab7783a2c1de

armbian/base/scripts/bbb-config.sh
```diff
@@ -125,7 +125,12 @@ case "${COMMAND}" in
 
                 redis_set "bitcoind:listen" "${ENABLE}"
                 generateConfig "bitcoin.conf.template"
-                systemctl restart bitcoind.service
+
+                # only restart bitcoind if system has been configured with setup routine
+                if [[ "$(redis_get 'base:setup')" -eq 1 ]]; then
```
0xB10C: Could use `base:bitcoind-services:enabled` instead of `base:setup`.

Stadicus: The key `base:bitcoind-services:enabled` should work as well, but I think coupling it to the finished setup is clearer. Not sure if we need `base:bitcoind-services:enabled` at all, at the momement it's only used to re-enable the services on every boot.

PR #251, opened 2019-11-07
Middleware: handle `mender-update` notification and queue if no client connected

0xB10C:
Events passed from the Middleware to the handlers are now typed. The type `Event` now includes a boolean which indicates if the event should be queued until a client (App in our case) is available.

Handling for the `mender-update` notification is add, that is send by a shell script after the mender update commit finishes.

For testing:

```
# success case:
echo '{"version": 1, "topic": "mender-update", "payload": {"success": true}}' >> /tmp/middleware-notification.pipe

# failure case:
echo '{"version": 1, "topic": "mender-update", "payload": {"success": false}}' >> /tmp/middleware-notification.pipe
```

### DIFF
--- a/middleware/src/handlers/handlers.go
+++ b/middleware/src/handlers/handlers.go
@@ -18,7 +18,7 @@ import (
 // Middleware provides an interface to the middleware package.
 type Middleware interface {
 	// Start triggers the main middleware event loop that emits events to be caught by the handlers.
-	Start() <-chan []byte
+	Start() <-chan Event
 
 	/* --- RPCs --- */
 	SystemEnv() rpcmessages.GetEnvResponse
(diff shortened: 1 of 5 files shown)


middleware/src/handlers/handlers.go
```diff
@@ -140,20 +155,31 @@ func (handlers *Handlers) versionHandler(w http.ResponseWriter, r *http.Request)
...
+
 	go server.Serve()
+
+	handlers.mu.Lock()
+	for _, event := range handlers.eventQueue {
+		for k := range handlers.clientsMap {
+			handlers.clientsMap[k] <- event.Identifier
+		}
+	}
+	handlers.mu.Unlock()
```
TheCharlatan: My understanding of go channels is still quite poor. I hope this works out nicely. Only fear I have here is that the channel will need to hold events until the Server is up and listening.

0xB10C: I'm not sure if I understand what you mean. Can you be a bit more specify?

TheCharlatan: It's probably fine, but the rpc server might not be running yet even after a connected client reacts to a notification with a rpc call.

middleware/src/middleware.go
```diff
@@ -211,9 +211,28 @@ func (middleware *Middleware) ipcNotificationLoop(reader *ipcnotification.Reader
 			log.Printf("Dropping IPC notification with unsupported version: %s\n", notification.String())
 		}
 
+		log.Printf("Received notification with topic '%s': %v\n", notification.Topic, notification.Payload)
+
 		switch notification.Topic {
 		case "sampletopic":
```
TheCharlatan: Can the sample be removed again now?

0xB10C:
Rebased and removed sample.

PR #252, opened 2019-11-07
update: send ipc notification to middleware

Stadicus:
In order to show a Update success/failure in the App, the Middleware needs to know if the update succeeded. This pull request implements IPC notification from update script to Middleware on success/failure

https://github.com/shiftdevices/bitbox-base-internal/issues/266

* the systemd-update-checks.sh script writes a success/failure msg
  into the middleware-notification.pipe, encoded in JSON

* must be merged and is already tested with https://github.com/digitalbitbox/bitbox-base/pull/251

### DIFF
--- a/armbian/base/scripts/systemd-update-checks.sh
+++ b/armbian/base/scripts/systemd-update-checks.sh
@@ -166,8 +166,14 @@ else
             else
                 echo "ERR: service verification try ${run} of 100 unsuccessful, falling back to previous Base image version"
                 redis_set "base:updating" 90
-                /opt/shift/scripts/bbb-cmd.sh base restart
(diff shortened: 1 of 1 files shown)


0xB10C:
Writing to a named pipe blocks until somebody on the other end reads. Should we add a `timeout` before the `echo` to not hang, when in fatal cases where middleware can not read from the pipe (i.e. not running or other critical error?).

_(`timeout` is a shell command)_

Stadicus:
Added `timeout` for 5 seconds for each IPC notification. Will add script function once this command is used more often.

Stadicus:
Added `|| true` to IPC notification call, so that the script does not fail completely if the named pipe is not available (best effort).

PR #254, opened 2019-11-08
add: FinalizeSetupWizard RPC

0xB10C:
The FinalizeSetupWizard RPC sets a Redis key indicating that the Base
has been setup, enables the Bitcoin services and starts the Bitcoin
services via the script added in #250.

### DIFF
--- a/middleware/src/middleware.go
+++ b/middleware/src/middleware.go
@@ -324,11 +324,6 @@ func (middleware *Middleware) BackupSysconfig() (response rpcmessages.ErrorRespo
 		}
 	}
 
-	//BackupSysconfig is the last call in the base setup. Once it is completed successfully, the setup is done.
-	err = middleware.redisClient.SetString(redis.BaseSetupDone, "1")
-	if err != nil {
-		log.Println("redis failed to set the base setup done flag. This is not a critical error, but the setup will be repeated again")
-	}
 	return rpcmessages.ErrorResponse{Success: true}
 }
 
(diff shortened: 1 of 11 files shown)


middleware/src/middleware.go
```diff
@@ -913,6 +908,43 @@ func (middleware *Middleware) SetLoginPassword(args rpcmessages.SetLoginPassword
 	}
 }
 
+// FinalizeSetupWizard finalizes the setup wizard by setting a redis key,
+// enabling bitcoin core and relates services and starting bitcoin code and
+// related services.
+func (middleware *Middleware) FinalizeSetupWizard() rpcmessages.ErrorResponse {
+	log.Println("Finalizing the setup wizard.")
+
+	err := middleware.redisClient.SetString(redis.BaseSetupDone, "1")
```
Stadicus: I think it would make sense to set the Redis key after the fact, to store success. If this RPC fails, it might not be called again although the services would stay disabled.

middleware/src/rpcmessages/errorcodes.go
```diff
@@ -139,6 +139,16 @@ const (
 	ErrorSetHostnameInvalidValue ErrorCode = "SET_HOSTNAME_INVALID_VALUE"
 )
 
+const (
+
+	/* bbb-systemctl.sh start-bitcoin-services
+	---------------------------------------*/
+
+	// ErrorSystemdServiceStartFailed is thrown when the systemd services can not
```
Stadicus: "when a systemd service cannot be started"? Could be used in other context as well and indicates usually a single failed attempt.

middleware/src/system/system_test.go
```diff
@@ -17,6 +17,7 @@ func TestSystem(t *testing.T) {
 	argumentMap["network"] = "testnet"
 	argumentMap["bbbConfigScript"] = "/home/bitcoin/config-script.sh"
 	argumentMap["bbbCmdScript"] = "/home/bitcoin/cmd-script.sh"
+	argumentMap["bbbSystemctlScript"] = "/home/bitcoin/bbb-systemctl.sh"
```
Stadicus: What are these paths? Is this the specific regtest setup? On the Base, there is no `/home/bitcoin` directory and these files are called via symlinks in `/usr/local/sbin/`

middleware/src/middleware.go
0xB10C: Agree

middleware/src/system/system_test.go
0xB10C: That are only test values to test the Getters. Could be a random string. might change them to avoid confusion.

middleware/src/middleware.go
```diff
@@ -913,6 +908,43 @@ func (middleware *Middleware) SetLoginPassword(args rpcmessages.SetLoginPassword
...
+		errorCode := handleBBBScriptErrorCode(out, err, []rpcmessages.ErrorCode{})
+
+		return rpcmessages.ErrorResponse{
+			Success: false,
+			Message: strings.Join(out, "\n"),
+			Code:    errorCode,
+		}
+	}
+
+	out, err = middleware.runBBBSystemctlScript([]string{"start-bitcoin-services"})
```
Tomasvrba: I think documentation comments here would be really nice here to quickly explain the difference between
`middleware.runBBBConfigScript([]string{"enable", "bitcoin_services"})` and ` middleware.runBBBSystemctlScript([]string{"start-bitcoin-services"})`, why both are needed in this order for anyone looking at this in the future.

0xB10C:
Rebased and addressed feedback.

PR #257, opened 2019-11-11
update 'build-ondevice' make target

Stadicus:
This pull request updates the build method to run the configuration script directly on a preinstalled Linux distribution. It's used mainly for testing of non-Armbian distributions.

**compile Go: use build instead of install in Docker**
https://github.com/shiftdevices/bitbox-base-internal/issues/326

Because:
* When cross-compiling the Go tools inside Docker, with the command
  'docker install', the binaries are put into the directory
  $GOPATH/bin/linux_arm64
* One goal is to run installation (incl. Go compilation) directly on
  the ARM device, to support other distributions. When not cross-
  compiling, the binaries are put into $GOPATH/bin/, failing the
  make process.

This commit:
* uses 'go build' instead of 'go install' to always put the binaries
  directly in the source directory and copy them from there.

**Update 'make build-ondevice'**
Because:
* Users should be able to run the BitBoxBase customization on a vanilla
  Linux installation, e.g. the Ayufan ARM distribution.
* For that to work, Go binaries need to be compiled first.
* Armbian specific changes are allowed to be skipped in that build mode

This commit:
* checks if Go tool binaries are available, or aborts
* adds some checks in customization script when running on device
* in prometheus-base.sh, check if Redis values like 'build:date' or
  'build:commit' are null and return 'n/a'.
* check if Redis factory settings are present, or abort
* outsource tasks not related to update out of systemd-update-checks.sh
  to systemd-startup-after-redis.sh

### DIFF
--- a/armbian/base/customize-armbian-rockpro64.sh
+++ b/armbian/base/customize-armbian-rockpro64.sh
@@ -299,6 +299,7 @@ mkdir -p /data/redis/
 chown -R redis:redis /data/redis/
 
 ### disable standard systemd unit
+systemctl stop redis-server.service || true
 systemctl disable redis-server.service
 systemctl mask redis-server.service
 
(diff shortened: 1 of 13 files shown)


armbian/base/customize-armbian-rockpro64.sh
```diff
@@ -476,7 +491,7 @@ redis-cli SET bitcoind:version "${BITCOIN_VERSION}"
 
 # LIGHTNING --------------------------------------------------------------------
 BIN_DEPS_TAG="v0.0.1-alpha"
-LIGHTNING_VERSION_BUILD="0.7.2.1"
+LIGHTNING_VERSION_BUILD="0.7.3"
```
0xB10C: Collides with https://github.com/digitalbitbox/bitbox-base/pull/256/commits/35bec85bcb9dc5f2a0ac52e7d1f298776bef642a from #256.  I'd prefer to not have it in here and only in #256.

Stadicus: Had to include it here as version 0.7.2.1  fails to build on the RockPro64 directly. Will merge https://github.com/digitalbitbox/bitbox-base/pull/256 first, remove that change, rebase and then merge this pull request.

PR #259, opened 2019-11-12
bbbmiddleware: don't use systemd EnvironmentFile

Stadicus:
After removing the bbbmiddleware.conf, the systemd unit must no longer try to load it.

This commit:
* fixes the systemd unit bbbmiddleware.service


Stadicus:
After digging deeper, there are a lot more scripts that require this template or the real file to be present. It's way easier to just create an empty template file for now, knowing that this is still the correct way to add configuration later.

Closing this pull-request, will add a new one re-adding the template and it's management.

PR #261, opened 2019-11-12
creating empty bbbmiddleware.conf

Stadicus:
related to https://github.com/digitalbitbox/bitbox-base/pull/255

Because:
* The bbbmiddleware.conf.template and the created file are used by many scripts.
* It was removed in pr255, which breaks both build and operations.

This commit:
* adds an empty bbbmiddleware.conf.template that can be recreated at will, sourced from systemd as environment file etc. without breaking the configuration.

Future configuration can be added here again.


Stadicus:
> grin

Nicest ACK I got ever! 😸

PR #266, opened 2019-11-14
Middleware: update GetBaseInfo 

0xB10C:
This is the second iteration of GetBaseInfo. 

- change: merge OnionPort & LocalPort into one
This change merges the MiddlewareLocalPort and MiddlewareOnionPort
from the GetBaseInfoResponse into one MiddlewarePort, since they are
the same for now.

- change: only return Onion if Tor is enabled
This change leaves the MiddlewareTorOnion emtpy ("") if Tor is not
enabled on the Base. This helps showing the correct values in the
node management.

- add: active lightning channels to BaseInfo
This adds the number of active lightning channels to the GetBaseInfo-
Response.

- add: isSSHPasswordLoginEnabled bool to BaseInfo
This adds a boolean indicating if SSH password login is enabled to
the GetBaseInfoResponse.

### DIFF
--- a/middleware/src/middleware.go
+++ b/middleware/src/middleware.go
@@ -1017,16 +1017,19 @@ func (middleware *Middleware) GetBaseInfo() rpcmessages.GetBaseInfoResponse {
 
 	middlewarePort := middleware.environment.GetMiddlewarePort()
 
-	middlewareTorOnion, err := middleware.redisClient.GetString(redis.MiddlewareOnion)
+	isTorEnabled, err := middleware.redisClient.GetBool(redis.TorEnabled)
 	if err != nil {
 		errResponse := middleware.redisClient.ConvertErrorToErrorResponse(err)
 		return rpcmessages.GetBaseInfoResponse{ErrorResponse: &errResponse}
 	}
 
-	isTorEnabled, err := middleware.redisClient.GetBool(redis.TorEnabled)
-	if err != nil {
-		errResponse := middleware.redisClient.ConvertErrorToErrorResponse(err)
-		return rpcmessages.GetBaseInfoResponse{ErrorResponse: &errResponse}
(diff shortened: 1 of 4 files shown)


middleware/src/middleware.go
```diff
@@ -1075,20 +1092,21 @@ func (middleware *Middleware) GetBaseInfo() rpcmessages.GetBaseInfoResponse {
...
-		ElectrsVersion:      electrsVersion,
+		Status:                    "-PLACEHOLDER-", // FIXME: This is a placeholder.
+		Hostname:                  hostname,
+		MiddlewareLocalIP:         middlewareIP,
+		MiddlewarePort:            middlewarePort,
+		MiddlewareTorOnion:        middlewareTorOnion,
+		IsTorEnabled:              isTorEnabled,
+		IsBitcoindListening:       isBitcoindListening,
+		LightningActiveChannels:   lightningActiveChannels,
+		IsSSHPasswordLoginEnabled: isSSHPasswordLoginEnabled,
```
Stadicus: Seems like `LightningActiveChannels` and `` are switched compared to the log output:

```
17:22:05 carol bbbmiddleware[2073]:  -PLACEHOLDER- carol 192.168.1.161 8845 5xla4g3brthhdyyg5ppxsvija3dysblv5tamealudpcu2lryj6k62oid.onion true true true 3 156599656448 502996557824 0.0.5 0.18.1 0.7.3 0.7.0}:
```

0xB10C: Good eye, but struct fields are ordered. Values can however be assigned to these fields in a different order.

Could clean up the ordering in the assignment, but otherwise everything works as expected.

Stadicus: Not an issue in this case. Just wanted to make sure that the `3` is really the Lightning number.

PR #269, opened 2019-11-16
scripts: remove base:bitcoind-services:enabled

Stadicus:
The Redis key 'base:bitcoind-services:enabled' was set after enabling the bitcoin-related services after the Setup Wizard. It was never checked anywhere else, though, all scripts are using 'base:setup' to decide whether to start these services.

This commit:
* removes key 'base:bitcoind-services:enabled'
* sets the key 'base:setup' to 1 also after manually enabling the bitcoin-related services with 'bbb-config.sh enable bitcoin-services' to ensure the services are started after IBD
* if the sync method should be changed after that (tor/clearnet/resync ect...), this can be triggered from the app.

### DIFF
--- a/armbian/base/config/redis/factorysettings.txt
+++ b/armbian/base/config/redis/factorysettings.txt
@@ -13,7 +13,6 @@ SET base:hostname bitbox-base
 SET base:update:allow-unsigned 0
 SET base:updating 0
 SET base:setup 0
-SET base:bitcoind-services:enabled 0
 
 SET middleware:passwordSetup 0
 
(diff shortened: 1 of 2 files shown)


Stadicus:
> This is already used by the middleware to finalize the setup wizard

Yes, that does not change. This RPC calls the following scripts:
*  `bbb-config.sh enable bitcoin_services`
*  `bbb-systemctl.sh start-bitcoin-services`

What I did change was the Redis key that is set: from `bitcoind-services` (not used anywhere else) to `base:setup`. No chanegs to the Middleware required, and now it also works if the scripts are run manually.

TL;DR: I think that change is correct.

PR #270, opened 2019-11-19
Service info fix

Tomasvrba:
- Move active lightning channels from base to service info
- return 0 instead of error for some unavailable values

Because:
* Certain values such as lightning and electrs blocks will not be
  available until bitcoin is fully synced. We still want to return
  useful information during the syncing process.
* Zeroes in this case make sense

This commit:
* return a 0 for bitcoindBlocks, bitcoindHeaders, lightningdBlocks,
  lightningActiveChannels, and electrsBlocks instead of an error
  when these are not available

### DIFF
--- a/middleware/src/util.go
+++ b/middleware/src/util.go
@@ -206,14 +206,12 @@ func (middleware *Middleware) didServiceInfoChange() (changed bool) {
 func (middleware *Middleware) getServiceInfo() rpcmessages.GetServiceInfoResponse {
 	bitcoindBlocks, err := middleware.prometheusClient.GetInt(prometheus.BitcoinBlockCount)
 	if err != nil {
-		errResponse := middleware.prometheusClient.ConvertErrorToErrorResponse(err)
-		return rpcmessages.GetServiceInfoResponse{ErrorResponse: &errResponse}
+		bitcoindBlocks = 0
 	}
 
 	bitcoindHeaders, err := middleware.prometheusClient.GetInt(prometheus.BitcoinHeaderCount)
 	if err != nil {
-		errResponse := middleware.prometheusClient.ConvertErrorToErrorResponse(err)
-		return rpcmessages.GetServiceInfoResponse{ErrorResponse: &errResponse}
+		bitcoindHeaders = 0
 	}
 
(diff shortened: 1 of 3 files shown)


middleware/src/util.go
```diff
@@ -206,14 +206,12 @@ func (middleware *Middleware) didServiceInfoChange() (changed bool) {
...
-		return rpcmessages.GetServiceInfoResponse{ErrorResponse: &errResponse}
+		bitcoindBlocks = 0
 	}
 
 	bitcoindHeaders, err := middleware.prometheusClient.GetInt(prometheus.BitcoinHeaderCount)
 	if err != nil {
-		errResponse := middleware.prometheusClient.ConvertErrorToErrorResponse(err)
-		return rpcmessages.GetServiceInfoResponse{ErrorResponse: &errResponse}
+		bitcoindHeaders = 0
 	}
```
0xB10C: Example for comment up top.

Tomasvrba:
> **Post merge NACK.**
> 
> > Certain values such as lightning and electrs blocks will not be
> > available until bitcoin is fully synced. We still want to return
> > useful information during the syncing process.
> 
> I think it's OK to return 0 for values that can't be fetched, but the implementation could be a lot cleaner.
> 
> Currently (with this PR merged) if an error occurs that's unrelated to the value not being available, we don't even get a log entry.
> 
> Why not add a `or vector(0)` to the query as in
> 
> ```go
> LightningActiveChannels     BasePrometheusQuery = "sum(lightning_peer_channels) or vector(0)"
> ```
> 
> which solves exactly this problem?

Yeah, we can definitely add `vector(0)` to the queries. That still doesn't solve insufficient error logging though no? The problem with the original implementation was that it was impossible to tell from the log entry which field had failed. All that you saw was a `PROMETHEUS_ERROR`, but no indication which query caused it. 

We ran into quite a lot of issues here with @Stadicus last week and spent hours troubleshooting which fields were not available at different times.

I ended up adding specific error logs to each field: https://github.com/Tomasvrba/bitbox-base/commit/39b855e02a554cfa20b18130693a4cd576019fce

Haven't merged it or opened a PR for that, but I'll do that now and we can discuss there.

PR #276, opened 2019-11-22
Pairing

benma:


middleware/src/noise/noise.go
```diff
@@ -21,25 +18,29 @@ import (
 const (
 	opICanHasHandShaek          = "h"
 	responseSuccess             = "\x00"
+	responseFailure             = "\x01"
```
0xB10C: Why have a both `responseNeedsPairing` `responseFailure` the same value of `\x01` @benma ?

benma: They are responses to different api calls. Maybe it's clearer to inline them where they are used.

@TheCharlatan can double check if the values make sense.

TheCharlatan: Doesn't really matter, since they respectively cover values expected by the client on the other side. Might be good to separate for clarities sake.

PR #277, opened 2019-11-22
middleware: return nil in case of error

benma:


Stadicus:
Some kind of reasoning for this change and a quick summary would be great. :-)

benma:
In case of error, the other return values are not used. Creating an instance is more overhead and syntactic noise, it is common to return the default values for them instead, which is `nil` for pointers.

PR #278, opened 2019-11-22
Fix / improve build process

Stadicus:
Armbian changed the build process and renamed their git branches. This pull-request fixes this, addresses some minor build improvements and extends the top-most `make` command.

**build fixes / improvements**
* switches to Armbian branch 'legacy' (fix)
* sets the .autosetup-ssd trigger as default and commits it also in overlayroot (bug)
* prompts for 'build.conf' when building ondevice
* optionally, run cleanup-loop-devices.sh before compilation
* source errorExit include in systemd-startup-checks.sh (bug)

**Makefile changes**
* extends 'make build-all' to include Mender postprocessing
* renames 'update' to 'update-all' and includes Mender postprocessing

### DIFF
--- a/armbian/armbian-build.sh
+++ b/armbian/armbian-build.sh
@@ -22,6 +22,11 @@ if ! [[ "${ACTION}" =~ ^(build|update|ondevice)$ ]]; then
 	exit 1
 fi
 
+# cleanup loop devices if trigger file present
+if [ -f .cleanup-loop-devices ]; then
+	../contrib/cleanup-loop-devices.sh
+fi
+
 case ${ACTION} in
 	build|update)
 		if ! command -v git >/dev/null 2>&1 || ! command -v docker >/dev/null 2>&1; then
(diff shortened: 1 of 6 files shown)


armbian/armbian-build.sh
```diff
@@ -44,8 +52,10 @@ case ${ACTION} in
 		cp -aR base/* armbian-build/userpatches/overlay/					# copy scripts and configuration items to overlay
 		cp -aR ../bin/go/* armbian-build/userpatches/overlay/bin/go			# copy additional software binaries to overlay
 
+		# TODO(Stadicus): pin specific tag
```
TheCharlatan: What tag are you referring to here?

Stadicus: We should checkout a specific release tag once things are more stable, e.g. `sunxi-5.3` (see https://github.com/armbian/build/releases). 

For that, we also need to follow "Using alternate armbian builder repos and branches"
https://docs.armbian.com/Developer-Guide_Build-Preparation/#using-alternate-armbian-builder-repos-and-branches

Not a priority, though.

PR #284, opened 2019-11-26
base/service-info: add error logging to each info field

Tomasvrba:
Adds error logging to each field in `BaseInfo` and `ServiceInfo` so that we can clearly tell which query had failed.


Tomasvrba:
This relates to discussion on the closed PR: https://github.com/digitalbitbox/bitbox-base/pull/270

0xB10C:
From the discussion in https://github.com/digitalbitbox/bitbox-base/pull/270#issuecomment-558519311: 
> All that you saw was a PROMETHEUS_ERROR, but no indication which query caused it.

I think adding the query to [this error (and in GetFloat() as well)](https://github.com/digitalbitbox/bitbox-base/blob/1907afbbd084f33eda6bfdfb1baec6ff45db6a94/middleware/src/prometheus/prometheus.go#L190) and then printing the error would be even more future proof (for Prometheus at least, not Redis).

PR #286, opened 2019-11-26
middleware-heartbeat: sends a heartbeat to the hsm every 5 seconds

Tomasvrba:
Initial implementation of the heartbeat sent by the middleware to the HSM.
Sends the heartbeat every 5 seconds with `IDLE` and `" "` (stateCode=0, descriptionCode=0)

Tested this on an HSM with a screen without the rest of the Base hardware (no redis). 

@0xB10C Here are the protobuf message definitions for how the codes are enumerated: https://github.com/digitalbitbox/bitbox02-api-go/blob/master/api/firmware/messages/bitboxbase.proto

So we can just have two redis keys, `stateCode` and `descriptionCode` and each of them storing an int from the enum range in the proto file.

### DIFF
--- a/middleware/src/middleware.go
+++ b/middleware/src/middleware.go
@@ -19,6 +19,7 @@ import (
 	"github.com/digitalbitbox/bitbox-base/middleware/src/redis"
 	"github.com/digitalbitbox/bitbox-base/middleware/src/rpcmessages"
 	"github.com/digitalbitbox/bitbox02-api-go/api/firmware"
+	"github.com/digitalbitbox/bitbox02-api-go/api/firmware/messages"
 	"github.com/digitalbitbox/bitbox02-api-go/util/semver"
 	"golang.org/x/crypto/bcrypt"
 )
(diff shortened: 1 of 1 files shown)


0xB10C:
I'll be cherry picking your commit and built on top of it.

middleware/src/middleware.go
```diff
@@ -181,8 +182,26 @@ func (middleware *Middleware) updateCheckLoop() {
...
 }
 
+// hsmHeartbeatLoop
+func (middleware *Middleware) hsmHeartbeatLoop() {
+	for {
+		// TODO(@0xB10C) fetch the `stateCode` and `descriptionCode` from redis keys set byt the supervisor
+		err := middleware.hsmFirmware.BitBoxBaseHeartbeat(messages.BitBoxBaseHeartbeatRequest_IDLE, messages.BitBoxBaseHeartbeatRequest_EMPTY)
+		if err != nil {
+			log.Printf("Received an error from the HSM: %s\n", err)
+			continue
```
0xB10C: Do we really want to continue here without waiting? If the next heartbeats fail as well we end up flooding the HSM with heartbeats.

Tomasvrba: Good call, I can add a 1-second sleep there before the continue.

0xB10C: Yeah, but should coordinate if we want to merge this PR or if we want to have it all in one.

0xB10C: no strong preference from my side.

Tomasvrba: Updated and rebased.
Fine with me either way as well.
I think in that case, let's just merge it.

PR #289, opened 2019-11-28
middleware/hsm: use blocking call over timeout (for now)

benma:
The serial package's max timeout is silently capped at 25.5s:

https://github.com/tarm/serial/blob/98f6abe2eb07edd42f6dfa2a934aea469acc29b7/serial.go#L158

We need longer for the blocking HSM operations to be able to finish.

We might add a shorter timeout and a repeater in the future, to still
abort after a (long) time, but unless we fully go async, all those
solutions are suboptimal.


NickeZ:
haha, intresting take on timeouts.

PR #293, opened 2019-12-01
docker: update dockerfile to use latest go1.13.4

x1ddos:
This is used to build tools/... and probably middleware/... I believe.
I suggest subscribing to https://groups.google.com/group/golang-announce to stay up to date with the regular and out of schedule releases.


x1ddos:
Sent a similar change to https://github.com/mendersoftware/mender-convert/pull/140.

PR #294, opened 2019-12-01
armbian: check out only one branch/commit

x1ddos:
I don't believe anyone needs full history of the armbian and
mender-convert repos. Last commit + branch should be enough.
One can always "git clone ..." manually with custom options.

This produces the same result without issues.


Stadicus:
Thanks, will also add this to the c-lightning repository checkout here in a follow-up pull request
https://github.com/digitalbitbox/bitbox-base/blob/d2d75f06f7fcde655209bbb11a01faafa3fb26c1/armbian/base/customize-armbian-rockpro64.sh#L520

Issue #292, opened 2019-12-01
Don't bundle "random" keys in armbian/base/config/ssh/authorized_keys


x1ddos:
It feels kind of weird that building the image by default includes "random" keys in https://github.com/digitalbitbox/bitbox-base/blob/d2d75f06f7fcde655209bbb11a01faafa3fb26c1/armbian/base/config/ssh/authorized_keys#L1-L4

Maybe a better option would be to add `armbian/base/config/ssh/authorized_keys` to .gitignore and document better how to add a key in the config? Say, somewhere around https://github.com/digitalbitbox/bitbox-base/blob/d2d75f06f7fcde655209bbb11a01faafa3fb26c1/armbian/base/build.conf#L23-L24

Stadicus:
This is for internal development purposes only and `off` by default.

But I agree, now that we reached a stage where we reliably can connect with the BitBoxApp and set the login password, we can remove that.

PR #295, opened 2019-12-03
remove SSH development keys

Stadicus:
fixes https://github.com/digitalbitbox/bitbox-base/issues/292

The SSH login can now be reliably set using the BitBoxApp, so untrusted SSH keys no longer need to be part of the official repo.

Users can still add their own keys in `armbian/base/config/ssh/authorized_keys`, which is excluded from Git versioning. To include these keys, also set `BASE_ADD_SSH_KEYS='true'` in `build.conf`.

This pull request
* removes the Shift ssh dev keys in `authorized_keys`
* adds `authorized_keys.readme` with additional information
* adds `authorized_keys` to `.gitignore`


armbian/base/config/ssh/authorized_keys.readme
```diff
@@ -0,0 +1,4 @@
+# For secure SSH login, add your SSH key to the file `authorized_keys` in this directory.
+# It is excluded from Git versioning control.
+#
+# See this introduction for additional information: <https://www.atlassian.com/git/tutorials/git-ssh>
```
x1ddos: Why not simply add the contents of this file above `BASE_ADD_SSH_KEYS` option in build.conf?
This file is much harder to discover compared to reading the doc comment in the config file.

Stadicus: I thougt about that but think this has its limits when working with multiple keys. Would work technically, I guess, but I'm not really comfortable with multi-line variables.

x1ddos: Multi-line variables? What I meant is this, in the `build.conf` file:

```
# Add SSH keys in ./config/ssh/authorized_keys file for remote login.
# The file is excluded from Git versioning control.
# See this introduction for additional information: <https://www.atlassian.com/git/tutorials/git-ssh>
#BASE_ADD_SSH_KEYS="false"
```

Stadicus: Ah, I read that wrongly and thought about adding the keys themselves to the build.conf. Good point, will add the info here.

Stadicus: User can now add own keys in `armbian/base/config/authorized_keys`, instructions are moved to `build.conf`

Issue #290, opened 2019-12-01
Mender's image won't boot from SD card


x1ddos:
Here's what I see from the serial output while trying to boot `bin/img-mender/0.1.0/BitBoxBase-v0.1.0-RockPro64.img` created from `bin/img-armbian/BitBoxBase_Armbian_RockPro64.img` with `make -C armbian mender-artefacts`, all built from commit d2d75f0 aka latest tag 0.1.0. 

```
vcc-phy-regulator@vcc_phy: ; enabling (ret: -38)
vdd-log@vdd_log: ; enabling (ret: -38)
MMC:   sdhci@fe330000: 0, dwmmc@fe320000: 1
Card did not respond to voltage select!
mmc_init: -95, time 19
*** Warning - No block device, using default environment

In:    serial@ff1a0000
Out:   serial@ff1a0000
Err:   serial@ff1a0000
Model: Pine64 RockPro64
normal boot
Net:   eth0: ethernet@fe300000
Hit any key to stop autoboot:  0
Card did not respond to voltage select!
mmc_init: -95, time 21
switch to partitions #0, OK
mmc1 is current device
Scanning mmc 1:1...
Found U-Boot script /boot.scr
reading /boot.scr
3136 bytes read in 3 ms (1020.5 KiB/s)
## Executing script at 00500000
Saving Environment to MMC...
Card did not respond to voltage select!
mmc_init: -95, time 20
!No block device
Boot script loaded from mmc 0:1
Card did not respond to voltage select!
mmc_init: -95, time 20
** Bad device mmc 0 **
Card did not respond to voltage select!
mmc_init: -95, time 21
"Synchronous Abort" handler, esr 0x96000010
ELR:     f7f6dd5c
LR:      f7f1b160
x0 : 0000000000000000 x1 : 0000000000000002
x2 : 0000000000000010 x3 : 0000000000000000
x4 : 00000000f7f81df0 x5 : 0000000000000002
x6 : 0000000000000008 x7 : 0000000000000044
x8 : 0000000000000001 x9 : 0000000000000008
x10: 00000000f5f0043f x11: 0000000000000006
x12: 000000000001869f x13: 00000000ffffffff
x14: 00000000f5f0297c x15: 00000000ffffffff
x16: 000000001e534356 x17: 00000000104080aa
x18: 00000000f5f0ee08 x19: 0000000000000000
x20: 00000000f5f34170 x21: 00000000f5f34100
[...]

PR #297, opened 2019-12-04
docs: improve DIY and Releases docs

Stadicus:
related: https://github.com/digitalbitbox/bitbox-base/issues/290

Because:
* It was not mentioned that only eMMC is supported and booting from a
  Mender MicroSD card does not work

This commit:
* adds this note about only supporting eMMC
* generally improves DIY and Releases docs


Stadicus:
Merging this docs update, contains only .md files.

PR #299, opened 2019-12-04
Middleware: Fix and mitigate switched parameters in Configuration factory function 

0xB10C:
This PR consists of two commits. 

The first fixes two switched parameters to the `NewConfiguration()` factory function. A Middleware without the fix the tires to HTTP GET update information from the `middlewarePort` (8845) and delivered the `imageUpdateInfoURL` as Middleware port to the App. This resulted in the App not getting notified about new updates. The bug was introduced in #285.

The second commit makes the `NewConfiguration()`function more developer friendly by requiring a struct with named fields as argument rather than just a sequence of (unnamed) parameters. For this an extra `Args` struct with public fields is used, which can be filled from the `main` package. The fields of the `Configuration` struct remain private and only accessible via Getters, because they should not be changeable while the Middleware is running.

### DIFF
--- a/middleware/src/configuration/configuration.go
+++ b/middleware/src/configuration/configuration.go
@@ -2,7 +2,37 @@
 // values to the Middleware.
 package configuration
 
+// Args has the same fields as the `Configuration` struct, but the fields in
+// `Args` are public. The struct is used as parameter to the `NewConfiguration()`
+// factory function. The struct needs public fields to be settable the `main`
+// package. However the `Configuration` can't have public fields, because the
+// fields should be **READ ONLY** while the Middleware is running. Fields should
+// only be accessible via defined Getter functions.
+//
+// Note: Go does not support named arguments for functions. While passing the
+// arguments to a function would work, it would be rather easy to mistakenly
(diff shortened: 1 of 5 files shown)


middleware/src/configuration/configuration.go
```diff
@@ -2,7 +2,37 @@
...
+// factory function. The struct needs public fields to be settable the `main`
+// package. However the `Configuration` can't have public fields, because the
+// fields should be **READ ONLY** while the Middleware is running. Fields should
+// only be accessible via defined Getter functions.
+//
+// Note: Go does not support named arguments for functions. While passing the
+// 			 arguments to a function would work, it would be rather easy to
+// 			 mistakenly switch e.g. two string parameters ending up with an invalid
+// 			 middleware configuration. Using the `Args` helps, because a Go struct
+//       can be initialized with named fields.
```
Stadicus: Mix of tab & space indent?

0xB10C: yeah probably. Will remove indent, as it's not really needed.

0xB10C:
Addressed indentation and rebased.

PR #301, opened 2019-12-05
Prometheus: log public network availability

Stadicus:
fixes https://github.com/shiftdevices/bitbox-base-internal/issues/370

Prometheus should log the availability of public internet without
leaking privacy information.

This pull requests queries an external host with a lot of general
traffic, Cloudflare, over Tor:

```
curl --socks5-hostname localhost:9050 1.1.1.1
```

If Tor is not active, it's impossible to ping an external host without
revealing the own ip address, but the solution can just "mingle in the
crowd", e.g. by `ping`ing Cloudflare, which should not be suspicious.

```
ping -c 1 1.1.1.1
```

These queries are run regularly from `prometheus-base.py`, so that the
result is collected by Prometheus and stored in its time-series database.
This helps with analyzing/debugging incidents, as the public network
availability can also be queried after the fact.

This commit:
* provides the new Prometheus metric 'base_internet_connectivity' that
  is 0 when OK, or an error code when NOT OK

### DIFF
--- a/armbian/base/scripts/prometheus-base.py
+++ b/armbian/base/scripts/prometheus-base.py
@@ -27,6 +27,7 @@ BASE_SYSTEMD_ELECTRS = Gauge("base_systemd_electrs", "Systemd unit status for El
 BASE_SYSTEMD_LIGHTNINGD = Gauge("base_systemd_lightningd", "Systemd unit status for c-lightning")
 BASE_SYSTEMD_PROMETHEUS = Gauge("base_systemd_prometheus", "Systemd unit status for Prometheus")
 BASE_SYSTEMD_GRAFANA = Gauge("base_systemd_grafana", "Systemd unit status for Grafana")
+BASE_INTERNET_CONNECTIVITY = Gauge("base_internet_connectivity", "Connectivity to public internet")
 
 r = redis.Redis(
     host='127.0.0.1',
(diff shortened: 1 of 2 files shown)


armbian/base/scripts/prometheus-base.py
```diff
@@ -80,6 +81,27 @@ def getSystemdStatus(unit):
...
 
+def getInternetConnectivity():
+    torEnabled = int(r.get('tor:base:enabled').decode("utf-8"))
+
+    try:
+        if torEnabled == 1:
+            print("Tor ok")
+            subprocess.check_output(["curl", "--socks5-hostname", "localhost:9050", "1.1.1.1"], shell=False, timeout=5, stderr=subprocess.STDOUT)
+        else:
+            print("Tor not ok")
```
0xB10C: Are L92 and L89 debug prints?

armbian/base/scripts/prometheus-base.py
```diff
@@ -80,6 +81,27 @@ def getSystemdStatus(unit):
...
+
+        return 0
+
+    except subprocess.TimeoutExpired as e:
+        print("getInternetConnectivity(): subprocess.TimeoutExpired; torEnabled", torEnabled)
+        return 1
+
+    except subprocess.CalledProcessError as e:
+        print("getInternetConnectivity(): subprocess.CalledProcessError (", e.returncode, "); torEnabled", torEnabled, e.output)
+        return e.returncode
```
0xB10C: > provides the new Prometheus metric 'base_internet_connectivity' that is 0 when OK, or an error code when NOT OK

I think that's a bit unintuitive to use 0 as OK and everything else as NOK. Either rename `base_internet_connectivity` to `base_no_internet_connectivity` or even better: return 1 if internet and 0 if no internet connectivity. Does curl or ping even return detailed info in the exit code?

armbian/base/scripts/prometheus-base.py
Stadicus: Yes, will remove them

armbian/base/scripts/prometheus-base.py
Stadicus: I agree that it's not very intuitive and will change it. 

I followed the logic of `getSystemdStatus`, which returns `0` if service is running. Maybe change this as well? Not sure, where that code is used elsewhere, in Supervisor or Middleware.

Stadicus: Commit https://github.com/digitalbitbox/bitbox-base/pull/301/commits/f64a68b121df419d17af090a5fba440940a39bfd changes logged values for both "network connectivity" and "systemd units" and updates the Grafana dashboard to reflect these changes.

Tomasvrba: Post-merge nACK from my side.
I really disagree that it's `unintuitive to use 0 as OK and everything else as NOK`, afaik that is the default convention in most systems. Definitely in all UNIX/POSIX systems and in C running on these systems. (https://en.wikipedia.org/wiki/Exit_status)
I'd say it's very unintuitive and confusing to flip these around and use 0 for not OK

Stadicus: As discussed offline, I agree if we look at the metric as a `exit-code`, this is how it was originally implemented. We switched to view the value more as `true`/`false` or `up`/`down`, where this logic makes more sense.

I'm happy to revisit that once we extend this, but it's not a critical part of the system just now.

PR #302, opened 2019-12-05
hsm-config: set hsm config on middleware start and SetHostname

Tomasvrba:
This commit:
Adds the `setHSMConfig()` function to call the HSM API's
`BitBoxBaseSetConfig()` on middleware startup and when the Base hostname
is changed using an RPC.
The IP and hostname are fetched from redis and prometheus respectively
whenever `setHSMConfig` is called.
For `statusLedMode` and `statusScreenMode` we currently use the default
`LED_ALWAYS `and `SCREEN_ALWAYS` until we have redis config for those values.

### DIFF
--- a/middleware/src/middleware.go
+++ b/middleware/src/middleware.go
@@ -200,6 +200,11 @@ func (middleware *Middleware) Start() <-chan handlers.Event {
 
 	go middleware.rpcLoop()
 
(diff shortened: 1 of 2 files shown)


Tomasvrba:
Tested: photo_2019-12-05_14-21-29

PR #303, opened 2019-12-05
HSM: add firmware to Base image

Stadicus:
fixes https://github.com/shiftdevices/bitbox-base-internal/issues/369

Currently, we update the HSM firmware together with the Base image. For the Middleware to update the HSM, the specific version of the HSM firmware needs to be included in the disk image.

This issue extends the build script to provide a specific release version in the disk image. The version is specified like `HSM_VERSION='4.3.0'`, that is then used to create the valid download URI like this:
```
https://github.com/digitalbitbox/bitbox02-firmware/releases/download/firmware-bitboxbase%2Fv4.3.0/firmware-bitboxbase.v4.3.0.signed.bin
```

It is important, that the `tag` of the release always follows this syntax: `firmware-bitboxbase/v4.3.0`
and the file is named like `firmware-bitboxbase.v4.3.0.signed.bin`.

The release binary file is then stored in the disk image:
```
/opt/shift/hsm/firmware-bitboxbase.signed.bin
```

The version is deliberately omited in the filename, as the Middleware reads the internal version number (of the monotonic counter) directly out of the file, and compares it with the currently installed firmware on the HSM.

### DIFF
--- a/armbian/base/customize-armbian-rockpro64.sh
+++ b/armbian/base/customize-armbian-rockpro64.sh
@@ -37,6 +37,8 @@ LIGHTNING_VERSION="0.7.3"
 ELECTRS_VERSION="0.7.0"
 BIN_DEPS_TAG='0.0.5'
 
+HSM_VERSION='4.3.0'
+
 PROMETHEUS_VERSION="2.11.1"
 PROMETHEUS_CHKSUM="33b4763032e7934870721ca3155a8ae0be6ed590af5e91bf4d2d4133a79e4548"
 NODE_EXPORTER_VERSION="0.18.1"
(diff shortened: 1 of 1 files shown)


armbian/base/config/redis/factorysettings.txt
```diff
@@ -14,6 +14,8 @@ SET base:update:allow-unsigned 0
 SET base:updating 0
 SET base:setup 0
 
+SET hsm:firmware:version xxx
```
benma: imho I would not store it in redis, this produces follow up effort in maintaining it when upgrading, being careful about that it is updated atomically with the firmware, etc. This is only good if we know concretely what to use it for.

Furthermore, the middleware knows the version of the HSM when it connects to it.

benma:
Otherwise it looks good to me, including the constant local filename (good thinking!).

armbian/base/config/redis/factorysettings.txt
Stadicus: Removed the Redis key/value pair.

Issue #291, opened 2019-12-01
Confusing option BASE_LOGINPW


x1ddos:
From the following: https://github.com/digitalbitbox/bitbox-base/blob/d2d75f06f7fcde655209bbb11a01faafa3fb26c1/armbian/base/build.conf#L26-L28

it would seem like this is setting the root password. However, it appears to be doing more than that. What it actually does is locking the user `base` if the root password wasn't set to a custom string in https://github.com/digitalbitbox/bitbox-base/blob/d2d75f06f7fcde655209bbb11a01faafa3fb26c1/armbian/base/customize-armbian-rockpro64.sh#L219-L223

When building the image, I set `BASE_ADD_SSH_KEYS=true` in `armbian/base/build-local.conf` and update `armbian/base/config/ssh/authorized_keys` to contain my ssh key.
I didn't assign any password to `BASE_LOGINPW` because all I want is to ssh with my key as `base` user, and this essentially locked me out.

So, it forces me to do the "unsafe" option, setting `BASE_LOGINPW` to a non-empty string, even though I won't use it.

Additionally, a message during the build is also confusing:

```
CONFIGURATION:
    USER / PASSWORD:    base / some-password
...
```

which would appear as if it was the `base` user password but in reality what it actually does later is:

```
+ BASE_LOGINPW_FINAL=some-password
+ echo root:some-password
+ chpasswd
```

I would suggest being more explicit and rename `BASE_LOGINPW` to something like `BASE_ROOT_PASSWORD` and don't conflate that option with the `base` user at all.

Stadicus:
This is not an easy topic and I won't claim we have the best solution yet, so thanks already for taking the time and going through with it.

For you to use all the features of the BitBoxBase image, you'll always need to use the free open-source BitBoxApp. The build options are mostly meant for development purposes, all advanced settings can later be set in the App. 

When building a **production image**, we don't want to know any password (be it user `root` or `base`) at all. This is why we are setting it to a random string. This is not enought, however, as we could easily learn it during the build process. Therefore, we also lock these users. This prevents anybody from logging in locally by connecting a keyboard and screen.
  * The BitBoxApp is required to configure all advanced settings for production images.
  * Users that want to gain local password login (with keyboard/screen) can "Set system password" in the App (equal to `BASE_LOGINPW`). The passwords for `root` and `base` are kept in sync, but the `root` user always stays locked. --> Local login with user `base` is possible.
[...]

PR #304, opened 2019-12-06
build: switch to build production image

Stadicus:
addresses https://github.com/digitalbitbox/bitbox-base/issues/291

The various build options make it difficult to spot if in total everyting is set correctly to build a production image. By adding the binary option `BASE_PRODUCTION_IMAGE` to build a PRODUCTION or DEV image, and setting all options to safe production defaults, configuration errors can be minimized.

This commit:
* adds the build option `BASE_PRODUCTION_IMAGE`
* sets all defaults to production values that can only be overwritten
  if `BASE_PRODUCTION_IMAGE=='false'`

### DIFF
--- a/armbian/base/build.conf
+++ b/armbian/base/build.conf
@@ -1,11 +1,24 @@
 # Configure Armbian image building process
-# --------------------------------------------------------------
+# ------------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


Stadicus:
This change requires updating `build-local.conf` for developers.

PR #314, opened 2019-12-12
hsm: add possibility to update HSM firmware from middleware

Tomasvrba:
This commit:
- Adds the `AvailableHSMVersion` Redis key: `hsm:firmware:version`
- Adds the `HSMUpdateAvailable()` method to the middleware which
compares the version of the running HSM firmware and the
AvailableHSMVersion Redis key
- If a new version is available, the `main()` function executes the
update via the hsm immediately after the middleware is started
- The signed firmware file is assumed to be in
`/opt/shift/hsm/firmware-bitboxbase.signed.bin` which is the default
value of the `hsmFirmwareFile` flag
- Together with the `updateHSMFirmware`, this can be used to force flash
firmware from a custom location
- If an update is successful, reboot back into the firmware via the
middleware so that the middleware has access to the new `hsmFirmware`


Tomasvrba:
@Stadicus We will also need to set the `AvailableHSMVersion` Redis key on build. I'll leave that to you?

middleware/cmd/middleware/main.go
```diff
@@ -31,9 +31,18 @@ func main() {
 	imageUpdateInfoURL := flag.String("updateinfourl", "https://shiftcrypto.ch/updates/base.json", "URL to query information about Base image updates from")
 	notificationNamedPipePath := flag.String("notificationNamedPipePath", "/tmp/middleware-notification.pipe", "Path where the Middleware creates a named pipe to receive notifications from other processes on the BitBoxBase")
 	hsmSerialPort := flag.String("hsmserialport", "/dev/ttyS0", "Serial port used to communicate with the HSM")
+	hsmFirmwareFile := flag.String("hsmfirmwarefile", "/opt/shift/hsm/firmware-bitboxbase.signed.bin", "Location of the signed HSM firmware binary")
```
benma: add to docstring that upgrade is automatically performed if the firmware file has a newer version.

middleware/src/hsm/hsm.go
```diff
@@ -203,3 +204,18 @@ func (hsm *HSM) InteractWithBootloader(f func(*bb02bootloader.Device)) error {
...
+// UpdateFirmware reboots into Bootloader and executes the firmware update
+func (hsm *HSM) UpdateFirmware(hsmFirmwareFile string) error {
+	hsmFirmwareBinary, err := ioutil.ReadFile(hsmFirmwareFile)
+	if err != nil {
+		return err
+	}
+	err = hsm.InteractWithBootloader(func(bootloader *bb02bootloader.Device) {
+		err = bootloader.UpgradeFirmware(hsmFirmwareBinary)
+	})
+	if err != nil {
```
benma: small nit: just `return err`

middleware/src/hsm/hsm.go
```diff
@@ -203,3 +204,18 @@ func (hsm *HSM) InteractWithBootloader(f func(*bb02bootloader.Device)) error {
...
 }
+
+// UpdateFirmware reboots into Bootloader and executes the firmware update
+func (hsm *HSM) UpdateFirmware(hsmFirmwareFile string) error {
+	hsmFirmwareBinary, err := ioutil.ReadFile(hsmFirmwareFile)
+	if err != nil {
+		return err
+	}
+	err = hsm.InteractWithBootloader(func(bootloader *bb02bootloader.Device) {
+		err = bootloader.UpgradeFirmware(hsmFirmwareBinary)
```
benma: it seems worthwhile to compare the monotonic version here again, as if the HSM boots in the bootloader from the start, the version is still checked.

alternatively, booting in booloader mode means something went wrong, so you could also just upgrade by default in that case.

as discussed offline, if this call fails, probably it's good to reboot the base and try again.

middleware/src/middleware.go
```diff
@@ -1171,3 +1172,32 @@ func (middleware *Middleware) VerifyAppMiddlewarePairing(channelHash []byte) (bo
 	}
 	return true, nil
 }
+
+// HSMUpdateAvailable checks the AvailableHSMVersion Redis and compares it to the running FW version
+func (middleware *Middleware) HSMUpdateAvailable(hsm *hsm.HSM) (bool, error) {
```
benma: hsm param unused? maybe better to pass `hsmFirmware` here instead

middleware/cmd/middleware/main.go
```diff
@@ -74,6 +83,25 @@ func main() {
...
+		updateAvailable, err := middleware.HSMUpdateAvailable(hsm)
+		if err != nil {
+			log.Printf("Failed to fetch HSM version/update information: %s. Proceeding normally", err)
+		}
+		if updateAvailable {
+			err = hsm.UpdateFirmware(*hsmFirmwareFile)
+			if err != nil {
+				log.Printf("Failed to update HSM firmware: %s", err)
+			} else {
+				err = middleware.BootHSMFirmware(hsm)
```
benma: This is a bit hard to follow. I'd just call `hsmFirmware, err := hsm.WaitForFirmware()` instead, and move everything up to line 52 to group all the HSM init related things.

middleware/src/hsm/hsm.go
```diff
@@ -203,3 +204,18 @@ func (hsm *HSM) InteractWithBootloader(f func(*bb02bootloader.Device)) error {
 	f(device)
 	return nil
 }
+
+// UpdateFirmware reboots into Bootloader and executes the firmware update
+func (hsm *HSM) UpdateFirmware(hsmFirmwareFile string) error {
```
benma: nit: `Upgrade` over `Update` as this word is also used in the bootloader and everywhere else.

middleware/src/middleware.go
Tomasvrba: Good catch, I was previously calling hsm.UpdateFirmware() from here, but then moved it out and forgot to remove this.

middleware/cmd/middleware/main.go
Tomasvrba: Moved to the middleware instead of main as discussed offline. Will finish the WIP tomorrow

middleware/src/hsm/hsm.go
Tomasvrba: Addressed in the following way:

initHSM tries to connect to the HSM firmware. On success, it checks if a firmware upgrade is available and tries to apply it.
On failing to connect to the firmware, it tries to connect to the bootloader and re-install the firmware.
If that also fails, we continue without the HSM

On purpose I am not rebooting automatically if any of this fails, as I don't want the Base to get stuck in a loop of failed upgrades. If a user see the upgrade has failed, they can reboot/re-plug manually.

If we decide to handle this automatically, we should store a persistent counter of failed upgrades and only try to reflash 2 or 3 times before informing the user that something has gone wrong.

Tomasvrba:
@benma Could you please take another quick look over my changes which address your review?

PR #321, opened 2019-12-30
base-version: fix bug where base version overwritten by hsm version

Tomasvrba:


Tomasvrba:
@Stadicus Correct me if I am wrong, but I believe that if this bug is live in user's beta Bases, they won't be able to do an OTA update unless we bump the json file version on the website to at least 4.4.1 instead of 0.1.1

PR #324, opened 2020-01-06
fix: set a default for the SUDO_USER variable

lausser:
If SUDO_USER is not set during a build, an error "unbound variable" is shown.

### DIFF
--- a/armbian/base/customize-armbian-rockpro64.sh
+++ b/armbian/base/customize-armbian-rockpro64.sh
@@ -420,7 +420,8 @@ fi
 if [[ "$BASE_OVERLAYROOT" == "true" ]]; then
   # allow missing files if build-ondevice (e.g. not Armbian distro)
   if [[ -f /etc/default/armbian-ramlog ]] || [[ "${BASE_BUILDMODE}" != "ondevice" ]]; then
-    sed -i '/ENABLED=/Ic\ENABLED=false' /etc/default/armbian-ramlog
+    -f /etc/default/armbian-ramlog && \
+      sed -i '/ENABLED=/Ic\ENABLED=false' /etc/default/armbian-ramlog
     sed -i 's/log.hdd/log/g' /etc/logrotate.conf
     importFile /etc/logrotate.d/rsyslog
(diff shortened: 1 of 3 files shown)


Stadicus:
Fix works, but we should also fix it here:
https://github.com/digitalbitbox/bitbox-base/blob/5ab3d5617a2a712aa6cd5b1f45a4b14e214f0ded/armbian/armbian-build.sh#L71

lausser:
I added this to my pull request.

lausser:
Oh, i accidentially pushed this commit, didn't know that it would extend the pending pull request. Should i find out how and create a separate pull reuest for the last commit?

armbian/base/customize-armbian-rockpro64.sh
```diff
@@ -420,7 +420,8 @@ fi
 if [[ "$BASE_OVERLAYROOT" == "true" ]]; then
   # allow missing files if build-ondevice (e.g. not Armbian distro)
   if [[ -f /etc/default/armbian-ramlog ]] || [[ "${BASE_BUILDMODE}" != "ondevice" ]]; then
```
Stadicus: This line already contains the check if the file to `sed` exists, which is only a requirement if the script is run on an existing distro with `make build-ondevice`. If the file does not existes when compiling Armbian from scratch, the script automatically fails due to `set -e` in the beginning.

armbian/base/customize-armbian-rockpro64.sh
```diff
@@ -448,7 +449,8 @@ generateConfig mender.conf.template # -->  /etc/mender/mender.conf
 
 ## configure swap file (disable Armbian zram, configure custom swapfile on ssd)
 if [[ -f /etc/default/armbian-zram-config ]] || [[ "${BASE_BUILDMODE}" != "ondevice" ]]; then
```
Stadicus: same here

lausser:
I didn't know that the script should intentionally break if this ramlog file is not there, (i ran into this error)
The commit is reverted now.

PR #326, opened 2020-01-16
build: use ayufan kernel for stability

Stadicus:
The standard Armbian image sufferst sometimes from device reboots or even kernel panics. Using the same image with the Ayufan kernel results in a much more stable image. If this is the case over a long period of time, we might switch to Ayufan completely, but that leads to major changes in the build system and mender post-processing.

For now, this commit replaces the Armbian kernel with the Ayufan kernel during build.

### DIFF
--- a/armbian/base/customize-armbian-rockpro64.sh
+++ b/armbian/base/customize-armbian-rockpro64.sh
@@ -500,6 +500,22 @@ curl --retry 5 -SLo "/opt/shift/hsm/firmware-bitboxbase.signed.bin" \
   "https://github.com/digitalbitbox/bitbox02-firmware/releases/download/firmware-bitboxbase%2Fv${HSM_VERSION}/firmware-bitboxbase.v${HSM_VERSION}.signed.bin"
 
 
+# AYUFAN KERNEL ---------------------------------------------------------------
+mkdir -p /usr/local/src/ayufan
+cd /usr/local/src/ayufan
+
+AYUFAN_KERNEL_VERSION="4.4.202-1237-rockchip-ayufan"
+AYUFAN_KERNEL_ID="gfd4492386213"
+
+curl --retry 5 -SLO "https://github.com/ayufan-rock64/linux-kernel/releases/download/${AYUFAN_KERNEL_VERSION}/linux-firmware-image-${AYUFAN_KERNEL_VERSION}-${AYUFAN_KERNEL_ID}_${AYUFAN_KERNEL_VERSION}_arm64.deb"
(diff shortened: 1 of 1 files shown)


Stadicus:
Tested building and running the resulting image, including Mender post-processing.

NickeZ:
How can you run things like `dpkg -i`? Are you building it in a virtual machine?

armbian/base/customize-armbian-rockpro64.sh
```diff
@@ -500,6 +500,22 @@ curl --retry 5 -SLo "/opt/shift/hsm/firmware-bitboxbase.signed.bin" \
...
+mkdir -p /usr/local/src/ayufan
+cd /usr/local/src/ayufan
+
+AYUFAN_KERNEL_VERSION="4.4.202-1237-rockchip-ayufan"
+AYUFAN_KERNEL_ID="gfd4492386213"
+
+curl --retry 5 -SLO "https://github.com/ayufan-rock64/linux-kernel/releases/download/${AYUFAN_KERNEL_VERSION}/linux-firmware-image-${AYUFAN_KERNEL_VERSION}-${AYUFAN_KERNEL_ID}_${AYUFAN_KERNEL_VERSION}_arm64.deb"
+curl --retry 5 -SLO "https://github.com/ayufan-rock64/linux-kernel/releases/download/${AYUFAN_KERNEL_VERSION}/linux-image-${AYUFAN_KERNEL_VERSION}-${AYUFAN_KERNEL_ID}_${AYUFAN_KERNEL_VERSION}_arm64.deb"
+
+dpkg --install linux-firmware-image-4.4.202-1237-rockchip-ayufan-gfd4492386213_4.4.202-1237-rockchip-ayufan_arm64.deb
```
NickeZ: Isn't the name of the file the same as in the curl command? In that case you could use a variable instead of repeating the same string.

Stadicus: Good catch, meant to replace that as well. One of the reasons to abstract it into the variables.

Stadicus: Fixed with force-pushed commit https://github.com/digitalbitbox/bitbox-base/pull/326/commits/b4ed5f9197944316c8359c11803d3937047383a4

Stadicus:
> How can you run things like `dpkg -i`? Are you building it in a virtual machine?

The build process runs in a Docker enviroment, but within that the whole script [customize-armbian-rockpro64.sh](https://github.com/digitalbitbox/bitbox-base/blob/master/armbian/base/customize-armbian-rockpro64.sh) is executed in a `chroot` environment, with virtually no limitation on linux commands.

Stadicus:
I put that pull-request on hold for now, as it is incompatible with `overlayroot` and caused other issues during extensive testing.

NickeZ:
> I put that pull-request on hold for now, as it is incompatible with `overlayroot`
How can it be incompatible with overlayroot? Does he compile the kernel without the overlay fs?

>  and caused other issues during extensive testing.
What kind of issues?

PR #331, opened 2020-01-22
Minor improvements

Stadicus:
* add `bash-completion` package on build
* allow `overlayroot-chroot` commands to fail with error logging
* use `apt-get` instead of `apt` in scripts (https://github.com/shiftdevices/bitbox-base-internal/issues/405)
* use `apt-get purge` instead of `apt-get remove` to also remove config files (related to https://github.com/shiftdevices/bitbox-base-internal/issues/404)

See commits for additional details.

### DIFF
--- a/armbian/base/customize-armbian-rockpro64.sh
+++ b/armbian/base/customize-armbian-rockpro64.sh
@@ -284,9 +284,9 @@ fi
 # SOFTWARE PACKAGE MGMT --------------------------------------------------------
 ## update system, force non-interactive commands
 
-apt -y update
-apt -y -q -o "Dpkg::Options::=--force-confdef" -o "Dpkg::Options::=--force-confold" upgrade
-apt -y --fix-broken install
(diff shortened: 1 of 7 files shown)


armbian/base/customize-armbian-rockpro64.sh
```diff
@@ -302,30 +302,30 @@ if [[ "${BASE_BUILDMODE}" != "ondevice" ]] && [[ "${BASE_MINIMAL}" == "true" ]];
 
   for pkg in $pkgToRemove
   do
-    apt -y remove "$pkg" || true
+    apt-get -y purge "$pkg" || true
```
TheCharlatan: What is the reason that all of them cannot be purged in one go?

Stadicus: See original issue
https://github.com/shiftdevices/bitbox-base-internal/issues/404#issuecomment-579718816

